PR #88, opened 2022-03-10
fix: allow authorization in Access-Control-Request-Headers

hughns:
@sandhose it might be the case that this header should not be allowed on all endpoints of the endpoints.

### DIFF
--- a/crates/handlers/src/oauth2/mod.rs
+++ b/crates/handlers/src/oauth2/mod.rs
@@ -15,6 +15,7 @@
 use std::sync::Arc;
 
 use hyper::Method;
+use hyper::header::AUTHORIZATION;
 use mas_config::{Encrypter, HttpConfig};
 use mas_jose::StaticKeystore;
 use mas_templates::Templates;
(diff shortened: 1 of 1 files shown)


crates/handlers/src/oauth2/mod.rs
```diff
@@ -59,7 +59,11 @@ pub fn filter(
         .unify()
         .or(introspection)
         .unify()
-        .with(cors().allow_methods([Method::POST, Method::GET]));
+        .with(
+            cors()
+                .allow_methods([Method::POST, Method::GET])
+                .allow_headers(["authorization"]),
```
sandhose, line 65: There are constants in hyper: <https://docs.rs/hyper/latest/hyper/header/constant.AUTHORIZATION.html>, so
```rust
use hyper::header::AUTHORIZATION;
```

and

```suggestion
                .allow_header(AUTHORIZATION),
```

sandhose (approved):
Nitpick, but otherwise LGTM

PR #173, opened 2022-05-04
Core page wording and flow

hughns:
Remove redundant buttons/links
Make wording more directive
Describe scope of access

### DIFF
--- a/crates/templates/src/res/account-base.html
+++ b/crates/templates/src/res/account-base.html
@@ -1,0 +1,45 @@
+{#
+Copyright 2021 The Matrix.org Foundation C.I.C.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+#}
+
+{% extends "base.html" %}
+
+{% block content %}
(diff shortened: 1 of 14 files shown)


crates/templates/src/res/account-base.html
```diff
@@ -0,0 +1,45 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+#}
+
+{% extends "base.html" %}
+
+{% block content %}
+<nav class="container mx-auto py-2 flex-initial flex items-center px-2" role="navigation" aria-label="main navigation">
```
sandhose, line 20: I think it makes more sense to have a partial/component and `include` it

crates/templates/src/res/pages/login.html
```diff
@@ -33,31 +33,33 @@
...
       {# TODO: errors #}
       {{ field::input(label="Username", name="username") }}
       {{ field::input(label="Password", name="password", type="password") }}
-      {{ button::button(text="Next") }}
-      {{ button::link_text(text="Create account", href=register_link) }}
+      {% if next and next.kind == "continue_authorization_grant" %}
+        <div class="grid grid-cols-2 gap-4">
+          {{ back_to_client::link(
+            text="Cancel",
+            class=button::text_class(),
```
sandhose, line 47: `class=button::plain_error_class()` would probably make it prettier

sandhose, line 47: `class=button::plain_error_class()` would probably make it prettier

sandhose (changes requested):

crates/templates/src/res/pages/login.html
hughns: I think I changed it from the solid red/alert block to the text as I felt it was distracting from the Next action.

How about this instead?

image

sandhose (approved):

PR #193, opened 2022-05-12
Legacy /login support

### LINKED ISSUE #21: Legacy login support via m.login.password
The auth service will need to handle the existing /login, /logout and /refresh APIs via the m.login.password login type.

sandhose:
Fixes #21 

This implements /login via m.login.password.
This is kinda temporary, since we probably want to support login via m.login.sso instead.

It also lacks a bunch of things:

 - support for /refresh (and token expiration)
 - support for /logout


sandhose:
- /logout is done 
 - /login does return an expiring access token and a refresh token if refresh_token=true in the request body
 - still misses /refresh

PR #221, opened 2022-05-31
Implementation of MSC3824 actions for compat

hughns:


crates/handlers/src/compat/login.rs
```diff
@@ -33,19 +33,32 @@ use thiserror::Error;
...
     #[serde(rename = "m.login.password")]
-    Password,
+    Password {
+        actions: Vec<Action>,
+    },
 
     #[serde(rename = "m.login.token")]
-    Token,
+    Token {
+        actions: Vec<Action>,
```
sandhose, line 54: Should this one have be empty array? If not, I'd either remove it here, or add
```
#[serde(skip_serializing_if = "Vec::is_empty")]
```
so it skips the field if the array is empty

sandhose (approved):
Nitpicks, but overall LGTM

crates/handlers/src/compat/login.rs
hughns: As per discussion on Matrix, I've removed `actions` for `m.login.token`.

PR #227, opened 2022-06-06
Build and publish docker image from main branch

hughns:


sandhose:
...I'm confused because it's already the case?

PR #248, opened 2022-06-14
feat: support for MSC3824 action param on SSO redirect

hughns:


sandhose (changes requested):
Pretty good overall! I think my main concern is around keeping the `action` in the PostAuthAction, see my comment in the patch

PR #276, opened 2022-06-29
Enforce redirect_uris verifications in client registration

sandhose:
See #275


sandhose:
The 'secure URI' and 'host match' rules are now configurable on an instance-basis. @hughns feel free to merge and deploy this

hughns (approved):

PR #285, opened 2022-07-01
Stop generating the device ID automatically

sandhose:


sandhose:
@hughns this should be fine to merge once clients are up to date. I already update hydrogen-web.

PR #320, opened 2022-08-01
Make PKCE implementation compliant with RFC7636

### LINKED ISSUE #316: Enforce compliant PKCE code_verifier lengths
https://datatracker.ietf.org/doc/html/rfc7636#section-4.1 says that valid code_verifier is 43 to 128 characters. Currently the length isn't validated.

The expected response for invalid length is:

`{"error":"invalid_grant","error_description":"PKCE invalid code verifier"}`

sandhose:
This checks for the PKCE code_verifier length as well as the characters
used. It also give better errors when the PKCE verifier is invalid.

Fixes #316

### DIFF
--- a/crates/data-model/src/oauth2/authorization_grant.rs
+++ b/crates/data-model/src/oauth2/authorization_grant.rs
@@ -24,7 +24,7 @@ use url::Url;
 use super::{client::Client, session::Session};
 use crate::{traits::StorageBackend, StorageBackendMarker};
 
-#[derive(Debug, Error)]
+#[derive(Debug, Error, PartialEq)]
 pub enum PkceVerificationError {
     #[error("code_verifier should be at least 43 characters long")]
     TooShort,
@@ -63,7 +63,7 @@ impl Pkce {
             return Err(PkceVerificationError::TooLong);
         }
 
-        if verifier
+        if !verifier
             .chars()
             .all(|c| c.is_ascii_alphanumeric() || c == '-' || c == '.' || c == '_' || c == '~')
         {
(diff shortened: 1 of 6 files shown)


crates/data-model/src/oauth2/authorization_grant.rs
```diff
@@ -24,6 +24,21 @@ use url::Url;
 use super::{client::Client, session::Session};
 use crate::{traits::StorageBackend, StorageBackendMarker};
 
+#[derive(Debug, Error)]
+pub enum PkceVerificationError {
+    #[error("code_verifier should be at least 43 characters long")]
```
sandhose, line 29: So here the `error_description` would look like that:

```
PKCE verification failed: code_verifier should be at least 43 characters long
```

crates/data-model/src/oauth2/authorization_grant.rs
```diff
@@ -39,9 +54,27 @@ impl Pkce {
...
 
-    #[must_use]
-    pub fn verify(&self, verifier: &str) -> bool {
-        self.challenge_method.verify(&self.challenge, verifier)
+    pub fn verify(&self, verifier: &str) -> Result<(), PkceVerificationError> {
+        if verifier.len() < 43 {
+            return Err(PkceVerificationError::TooShort);
+        }
+
+        if verifier.len() > 43 {
```
hughns, line 62:
```suggestion
        if verifier.len() > 128 {
```

hughns (changes requested):
Condition for max length is wrong. Otherwise looks good.

sandhose:
Actually the allowed characters check was also buggy. Fixed it and added a bunch of unit tests

hughns (approved):

Issue #313, opened 2022-07-23
`oauth2-types` for clients


zecakeh:
I'm working on [adding support for OIDC in the matrix-rust-sdk](https://github.com/matrix-org/matrix-rust-sdk/issues/859), at least as a PoC for now (so clients like ElementX could test it).

I've tested locally based on two crates:

- [openidconnect](https://crates.io/crates/openidconnect): it seems to work fine apart from a little incompatibility where it expects the OP to return the client metadata when dynamic registration is successful. It's a SHOULD in the [OpenID Connect Dynamic Client Registration spec](https://openid.net/specs/openid-connect-registration-1_0.html#RegistrationResponse), so the crate should probably not require it.

- this project's oauth2-types: I had to make a few modifications to be able to build some types that had private fields, but other than that it's pretty usable.

I preferred to work with oauth2-types even if everything needs to be done manually.

Is it planned for oauth2-types to have first class client support? I already see in the docs that it's mentioned it might be worth publishing this crate.

By first-class I mean integrating with [http](https://crates.io/crates/openidconnect) to build requests or extract responses for example. There could also be `client` vs `server` features, to improve compile time by only compiling what is necessary (e.g. for a client, a request doesn't need to be deserialized, and a response doesn't need to be serialized).

sandhose:
> I'm working on https://github.com/matrix-org/matrix-rust-sdk/issues/859, at least as a PoC for now (so clients like ElementX could test it).

That is really good to hear! I planned to do that anyway once I'm back (currently out of office until the 1st of august), so I would be glad to help you on that

> Is it planned for oauth2-types to have first class client support? I already see in the docs that it's mentioned it might be worth publishing this crate.

We need OIDC 'client' features in MAS anyway at some point, because we will need to support upstream OPs (see #19).


> By first-class I mean integrating with [http](https://crates.io/crates/openidconnect) to build requests or extract responses for example.

For doing HTTP request, I would like to leverage [`tower`](https://docs.rs/tower). You would provide a `Service<http::Request, Response = http::Response>`, with whatever layer you want (see for example what I'm doing [here](https://github.com/matrix-org/matrix-authentication-service/blob/ba6a382f2cde944fb7235c7969970958a2a695d9/crates/http/src/layers/client.rs#L68-L83) for the HTTP client I'm using for requesting clients JWKS).

---

[...]

PR #321, opened 2022-08-02
Allow to validate provider metadata

zecakeh:
Validation is done according to [OpenID Connect Discovery 1.0 spec](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata).

`Metadata` is renamed to `ProviderMetadata`, for the parallel with `ClientMetadata`.

`ProviderMetadata` has accessors for fields that have a default value and  `VerifiedProviderMetadata` has methods to access required fields directly.

Finally, it implements `Deserialize` to be able to use it for clients.

I couldn't find where the `prompt_values_supported` field comes from, since it's not in the IANA registry linked in the description, it might be worth it to add a link in the docs.

Part of #313.

### DIFF
--- a/crates/oauth2-types/src/oidc.rs
+++ b/crates/oauth2-types/src/oidc.rs
@@ -224,7 +224,7 @@ pub struct Metadata {
     pub require_signed_request_object: Option<bool>,
 
     /// URL of the authorization server's pushed authorization request endpoint.
-    pub pushed_authorization_request_endpoint: Option<bool>,
+    pub pushed_authorization_request_endpoint: Option<Url>,
 
     /// Indicates whether the authorization server accepts authorization
     /// requests only via PAR.
--- a/crates/oauth2-types/src/oidc.rs
+++ b/crates/oauth2-types/src/oidc.rs
@@ -14,6 +14,7 @@
 
 use std::ops::Deref;
 
+use language_tags::LanguageTag;
 use mas_iana::{
     jose::{JsonWebEncryptionAlg, JsonWebEncryptionEnc, JsonWebSignatureAlg},
     oauth::{
@@ -114,11 +115,10 @@ pub struct ProviderMetadata {
     /// might want or need to know when using the authorization server.
     pub service_documentation: Option<Url>,
 
-    // TODO: type
     /// Languages and scripts supported for the user interface, represented as a
     /// JSON array of language tag values from BCP 47. If omitted, the set of
     /// supported languages and scripts is unspecified.
-    pub ui_locales_supported: Option<Vec<String>>,
+    pub ui_locales_supported: Option<Vec<LanguageTag>>,
 
     /// URL that the authorization server provides to the person registering the
     /// client to read about the authorization server's requirements on how the
@@ -215,10 +215,9 @@ pub struct ProviderMetadata {
     /// OpenID Provider MAY be able to supply values for.
     pub claims_supported: Option<Vec<String>>,
 
-    // TODO: type
     /// Languages and scripts supported for values in Claims being returned,
     /// represented as a JSON array of BCP 47 language tag values.
-    pub claims_locales_supported: Option<Vec<String>>,
+    pub claims_locales_supported: Option<Vec<LanguageTag>>,
 
     /// Boolean value specifying whether the OP supports use of the "claims"
     /// parameter.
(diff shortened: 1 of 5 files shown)


crates/oauth2-types/src/oidc.rs
```diff
@@ -233,3 +250,541 @@ pub struct Metadata {
...
+    #[error("supported subject types are missing")]
+    MissingSubjectTypesSupported,
+
+    /// The supported ID token signing algorithm values are missing.
+    #[error("supported ID token signing algorithm values are missing")]
+    MissingIdTokenSigningAlgValuesSupported,
+
+    /// The URL of the given field doesn't use a `https` scheme.
+    #[error("{0}'s URL doesn't use a https scheme: {1}")]
+    UrlNonHttpsScheme(String, Url),
```
sandhose, line 657: Those strings are always known at compile time, right? They could be `&'static str` then, avoiding some `to_owned` calls here and there

crates/oauth2-types/src/oidc.rs
```diff
@@ -192,44 +287,588 @@ pub struct Metadata {
...
+
+/// Possible extra restrictions on a URL.
+#[derive(Debug, PartialEq, Eq)]
+enum ExtraUrlRestrictions {
+    /// The URL must not contain a fragment.
+    NoFragment,
+
+    /// The URL must not contain a query or a fragment.
+    NoQueryOrFragment,
+}
```
sandhose, line 805: What about including a `None` possible value here, to avoid the option? It would make the `validate_url` logic simpler

Something like

```rust
/// Possible extra restrictions on a URL.
#[derive(Debug, PartialEq, Eq, Copy)]
enum UrlRestrictions {
    None,

    /// The URL must not contain a fragment.
    NoFragment,

    /// The URL must not contain a query or a fragment.
    NoQueryOrFragment,
}

impl UrlRestrictions {
    const fn can_have_fragment(self) -> bool {
       self == Self::None
    }

    const fn can_have_query(self) -> bool {
       self != Self::NoQueryFragment
    }
}

fn validate_url(
     field: &'static str,
     url: &Url,
     restrictions: UrlRestrictions,
 ) -> Result<(), ProviderMetadataVerificationError> {
     if url.scheme() != "https" {
         return Err(ProviderMetadataVerificationError::UrlNonHttpsScheme(
             field,
             url.clone(),
         ));
     }

 
     if !restrictions.can_have_query() && url.query().is_some() {
         return Err(ProviderMetadataVerificationError::UrlWithQuery(
             field,
             url.clone(),
         ));
     }

     if !restrictions.can_have_fragment() && url.fragment().is_some()
     {
         return Err(ProviderMetadataVerificationError::UrlWithFragment(
             field,
             url.clone(),
         ));
     }

     Ok(())
 }

```

crates/oauth2-types/src/oidc.rs
```diff
@@ -192,44 +287,588 @@ pub struct Metadata {
...
+/// authentication methods.
+///
+/// The restrictions are:
+/// - The algorithm values must not contain `none`,
+/// - If the `client_secret_jwt` or `private_key_jwt` authentication methods are
+///   supported, the values must be present.
+fn validate_signing_alg_values_supported(
+    endpoint: &'static str,
+    values: Option<&[JsonWebSignatureAlg]>,
+    methods: Option<&[OAuthClientAuthenticationMethod]>,
```
sandhose, line 855: Here your could take `&Option<T>` directly instead of `Option<&T>` to avoid having to call `Option::as_deref()` everywhere

sandhose (changes requested):
I love this! Would you mind writing some tests for the validation? Maybe not every single field, but common things like required fields, and complex things like verifying the `*_auth_signing_alg_values_supported` and `*_auth_methods_supported` interactions.

zecakeh:
I'm wondering if we should return an array of errors, since there can be so many validation errors and it could be annoying to fix them one by one.

On the other end, all the requirements are in the docs now!

sandhose (approved):
Those tests are quite extensive, that's very nice to see!

I guess the validation could return a list of errors, but for now I think this is good enough 👍 
I'll let the CI pass (I'm getting some clippy warnings locally you'd probably have to fix) and merge once it's green!

zecakeh:
Yeah sorry I forgot to run cargo clippy. I'm used to getting clippy errors in the IDE along the format errors but it's not the case by default in this project. I'll have to look at the config.

sandhose:
Hah, forgot to merge it, sorry about that.

PR #334, opened 2022-08-04
Database testing

sandhose:
This adds tests which depend on a postgres database.
It leverages the new `sqlx::test` introduced in sqlx 0.6.1, which creates, migrate and teardown a database for each test.

This PR does 3 things:

 - setup GitHub Actions to have a PostgreSQL instance when running tests
 - adds a first simple database test in `mas-storage`, which registers a user, tries to login, etc.
 - adds a helper to create an `axum::Router` used in tests, which helps writing tests in `mas-handlers`

Note that this also enables the `wasmtime` compilation cache to speed up tests (else we compile the WASM module on each test in `mas-handlers`, which takes about a second)

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -259,12 +259,20 @@ jobs:
             target
           key: ${{ runner.os }}-cargo-build-${{ steps.toolchain.outputs.rustc_hash }}-${{ hashFiles('**/Cargo.lock') }}
 
+      - name: Start PostgreSQL database
+        run: |
+          sudo systemctl start postgresql.service
+          pg_isready
+          sudo -u postgres psql --command="CREATE USER test PASSWORD 'test'" --command="\du"
+
       - name: Test
         id: test
         uses: actions-rs/cargo@v1
         with:
           command: test
           args: --offline --workspace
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yaml
```diff
@@ -265,6 +280,9 @@ jobs:
         with:
           command: test
           args: --offline --workspace
+        env:
+          DATABASE_URL: postgresql://postgres:postgres@localhost/postgres
+          SQLX_OFFLINE: '1'
```
sandhose, line 285: This is needed, else sqlx will try to type check the queries against the live database

crates/policy/src/lib.rs
```diff
@@ -67,6 +70,9 @@ impl PolicyFactory {
         let mut config = Config::default();
         config.async_support(true);
         config.cranelift_opt_level(wasmtime::OptLevel::Speed);
+        config
+            .cache_config_load_default()
+            .map_err(LoadError::CacheSetup)?;
```
sandhose, line 75: This loads the default cache config, which should help with speeding up the tests.

On my laptop, without cache it looks like it adds ~1s to each test

PR #340, opened 2022-08-08
Disable wasmtime cache in the Docker image

sandhose:
This disables the wasmtime cache in the Docker image because

 1. it probably would never be used (unless you mounted `/home/nonroot/.cache/wasmtime` to a persistent place)
 2. makes it crash if it can't write to the default cache location (so if you run the image with `--read-only` for example)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -132,12 +132,22 @@ RUN mv target/$(/docker-arch-to-rust-target.sh "${TARGETPLATFORM}")/release/mas-
 
 ## Runtime stage, debug variant ##
 FROM --platform=${TARGETPLATFORM} gcr.io/distroless/cc-debian${DEBIAN_VERSION}:debug-nonroot AS debug
+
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -132,12 +132,22 @@ RUN mv target/$(/docker-arch-to-rust-target.sh "${TARGETPLATFORM}")/release/mas-
 
 ## Runtime stage, debug variant ##
 FROM --platform=${TARGETPLATFORM} gcr.io/distroless/cc-debian${DEBIAN_VERSION}:debug-nonroot AS debug
+
+# Inject a wasmtime config which disables cache to avoid issues running with a read-only root filesystem
+ENV XDG_CONFIG_HOME=/etc
+COPY ./misc/wasmtime-config.toml /etc/wasmtime/config.toml
+
```
sandhose, line 139: We could also push the config to `/home/nonroot/.config/wasmtime` instead and not set the `XDG_CONFIG_HOME`, but at least here it works even if it runs under another user

PR #352, opened 2022-08-12
Better Tower layers

sandhose:
@zecakeh I'm trying to improve the tower layers in mas-http ; I'll also gate some parts behind features so you don't get unnecessary dependencies when using it in the OIDC client

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -108,9 +108,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.59"
+version = "1.0.62"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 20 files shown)


zecakeh (approved):
These are mostly nitpicks, otherwise it looks good.

Do we plan to provide a default mapper for `oauth2_types::ClientError`?

Another nitpick, in the `ClientLayer` there is this comment:

https://github.com/matrix-org/matrix-authentication-service/blob/61f4287ad12abd04bcb1ecc64c93d604483d6a6f/crates/http/src/layers/client.rs#L63-L67

It is not true for the `DecompressionLayer` that just changes the response type. Also, if we want, the `tower-http` crate has a [`TimeoutLayer`](https://docs.rs/tower-http/latest/tower_http/timeout/index.html) that doesn't change the error, it just returns a `408 Request Timeout` HTTP status code.

PR #354, opened 2022-08-16
Allow to validate client metadata

zecakeh:
According to OpenID Connect Dynamic Client Registration Spec 1.0.

Introduce `VerifiedClientMetadata`.

We should probably forward the validation errors to the user instead of using always the same error description?

This doesn't store localized metadata in the database.

Part of #313.


sandhose (changes requested):
Looks really good!

Just a few comments on the deserialisation. Could you add a test which covers serialisation and deserialisation of the `ClientMetadata`?

Something like this:

```rust

    #[test]
    fn deserialize_localized_fields() {
        let metadata = serde_json::json!({
            "redirect_uris": ["http://localhost/oidc"],
            "client_name": "Postbox",
            "client_name#fr": "Boîte à lettres",
        });

        let metadata: ClientMetadata = serde_json::from_value(metadata).unwrap();
        let name = metadata.client_name.unwrap();
        assert_eq!(name.non_localized(), "Postbox");
        assert_eq!(
            name.get(Some(&LanguageTag::from_str("fr").unwrap()))
                .unwrap(),
            "Boîte à lettres"
        );
    }
```

zecakeh:
Thanks for the review. Are the (de)serialization tests enough like that or do you want to see more fields covered?

crates/handlers/src/oauth2/registration.rs
```diff
@@ -111,55 +123,21 @@ pub(crate) async fn post(
...
+    let res = policy.evaluate_client_registration(&metadata).await?;
     if !res.valid() {
         return Err(RouteError::PolicyDenied(res.violations));
     }
 
+    // Contacts was checked by the policy
+    let contacts = match metadata.contacts.as_deref() {
+        Some(contacts) => contacts,
+        None => unreachable!(),
+    };
```
sandhose, line 139: Why isn't this one a getter/method in `VerifiedClientMetadata` like the other ones?

sandhose (changes requested):
One last thing, otherwise we're good!

crates/handlers/src/oauth2/registration.rs
zecakeh: Because it's not mandatory in OpenID Connect, so it's only enforced by the OPA policy.

sandhose: Makes sense! :)
Although, the HS admin might run the server with a policy which might not require the contacts field, so I'd rather default to an empty slice here instead of crashing

zecakeh: That's right I was wondering if the policy could be changed and I forgot to ask you.

sandhose (approved):

PR #375, opened 2022-08-19
Make HTTP request layers convert to `Bytes`

zecakeh:
Add layer to convert a request's `Bytes` to `Body`.

Renames `BodyToBytes` to `BodyToBytesResponse` to make the difference with `BytesToBodyRequest` clearer.

I didn't realize when reviewing #352 that the layers were returning a `Body`, so they didn't work when trying to rebase #347.

Like that we can use both services that use `Bytes` and `Body` easily.

### DIFF
--- a/crates/http/src/layers/bytes_to_body_request.rs
+++ b/crates/http/src/layers/bytes_to_body_request.rs
@@ -1,0 +1,66 @@
+// Copyright 2022 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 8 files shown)


crates/http/src/layers/bytes_to_body_request.rs
```diff
@@ -0,0 +1,71 @@
...
+        let (parts, body) = request.into_parts();
+        let body = Full::new(body);
+
+        let request = Request::from_parts(parts, body);
+
+        let inner = self.inner.call(request);
+
+        let fut = async { inner.await };
+
+        Box::pin(fut)
```
sandhose, line 58: You probably don't have to box the future here, since we're not mapping anything after.

You could set the associated type `type Future = S::Future;` and then just return `self.inner.call(request)`

sandhose (approved):
I'm not 100% sure it's really necessary, probably that having something more generic for the HTTP client #347 could solve this, but if that helps you for now, sure :)

Small nitpick over an unnecessary boxing of a future, but other than that, LGTM

zecakeh:
> I'm not 100% sure it's really necessary, probably that having something more generic for the HTTP client #347 could solve this, but if that helps you for now, sure :)

I'd be fine with using something more generic for the HTTP client but I couldn't find a common trait between `hyper` and `reqwest`. It seems a waste to make it generic over `Body` because generally we convert it from/into a `Bytes`, and that's already a type accepted/returned by `reqwest`.

PR #383, opened 2022-08-24
Use an enum for client error codes

zecakeh:
Replace the `ClientError` constants with `From<ClientErrorCode>`.


sandhose (approved):
LGTM!

PR #384, opened 2022-08-24
Use an enum for scope tokens

zecakeh:


crates/oauth2-types/src/scope.rs
```diff
@@ -14,35 +14,86 @@
...
 #[error("Invalid scope format")]
 pub struct InvalidScope;
 
-#[derive(Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash)]
-pub struct ScopeToken(Cow<'static, str>);
+/// Tokens to define the scope of an access token or to request specific claims.
+#[derive(
+    Debug, Clone, PartialEq, Eq, PartialOrd, Ord, Hash, SerializeDisplay, DeserializeFromStr,
+)]
+pub enum ScopeToken {
```
sandhose, line 32: I don't think that should live here. This crate is supposed to be usable in non-Matrix context, so having the Matrix device scope as a enum member here feels weird.

I'd rather keep the `ScopeToken` here as is, and have this type live somewhere else, with an `impl From<ScopeToken> for WhateverScopeType` and `impl TryFrom<WhateverScopeType> for ScopeToken`. What do you think?

sandhose (changes requested):

crates/oauth2-types/src/scope.rs
zecakeh: Sure, should it just be a new type in the client crate?

zecakeh:
This will be moved into #347 instead.

PR #389, opened 2022-09-01
Improve oauth2-types requests

zecakeh:
Make them usable from the client side and improve the docs.

### DIFF
--- a/crates/oauth2-types/src/requests.rs
+++ b/crates/oauth2-types/src/requests.rs
@@ -31,6 +31,10 @@ use crate::scope::Scope;
 
 // ref: https://www.iana.org/assignments/oauth-parameters/oauth-parameters.xhtml
 
+/// The mechanism to be used for returning Authorization Response parameters
+/// from the Authorization Endpoint.
+///
+/// Defined in [OAuth 2.0 Multiple Response Type Encoding Practices](https://openid.net/specs/oauth-v2-multiple-response-types-1_0.html#ResponseModes).
 #[derive(
     Debug,
     Hash,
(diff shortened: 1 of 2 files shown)


crates/oauth2-types/src/requests.rs
```diff
@@ -91,90 +135,226 @@ pub enum Display {
...
+/// [Token Endpoint]: https://www.rfc-editor.org/rfc/rfc6749#section-3.2
+/// [Authorization Code]: https://www.rfc-editor.org/rfc/rfc6749#section-4.1
 #[skip_serializing_none]
-#[derive(Serialize, Deserialize, Debug, PartialEq, Eq)]
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 pub struct AuthorizationCodeGrant {
+    /// The authorization code that was returned from the authorization
+    /// endpoint.
     pub code: String,
-    #[serde(default)]
```
sandhose, line 156: Haven't tested, but there are multiple places where you deleted `#[serde(default)]` annotations. Isn't that a problem when deserialising, and the field is absent?

crates/oauth2-types/src/requests.rs
```diff
@@ -191,40 +371,62 @@ pub struct ClientCredentialsGrant {
...
+///
+/// [Token Endpoint]: https://www.rfc-editor.org/rfc/rfc6749#section-3.2
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
 #[serde(tag = "grant_type", rename_all = "snake_case")]
 pub enum AccessTokenRequest {
     AuthorizationCode(AuthorizationCodeGrant),
     RefreshToken(RefreshTokenGrant),
     ClientCredentials(ClientCredentialsGrant),
-    #[serde(skip_deserializing, other)]
+    #[serde(skip_serializing, skip_deserializing, other)]
```
sandhose, line 396:
```suggestion
    #[serde(skip, other)]
```

sandhose (changes requested):

crates/oauth2-types/src/requests.rs
zecakeh: The attribute is not needed on `Option`s. It defaults to `None` automatically.

sandhose (approved):

PR #390, opened 2022-09-02
Remove support for the `token` response type

sandhose:
This removes support for the `token` response type, since it's highly discouraged and deprecated in OAuth 2.1.


crates/handlers/src/oauth2/discovery.rs
```diff
@@ -74,11 +74,7 @@ pub(crate) async fn get(
 
     let response_types_supported = Some(vec![
         OAuthAuthorizationEndpointResponseType::Code,
-        OAuthAuthorizationEndpointResponseType::Token,
         OAuthAuthorizationEndpointResponseType::IdToken,
-        OAuthAuthorizationEndpointResponseType::CodeToken,
-        OAuthAuthorizationEndpointResponseType::CodeIdToken,
```
sandhose, line 80: We apparently had `code id_token` twice here

zecakeh (approved):
:tada:

PR #408, opened 2022-09-05
Allow to get verified provider metadata with minimal checks

zecakeh:


zecakeh:
Oops I actually forgot that we're only suppose to check for the required fields.

sandhose (approved):

PR #414, opened 2022-09-06
Move claim hash token function to mas-jose crate

zecakeh:
Change the hash function according to the signature algorithm,
according to the OpendID Connect spec.


crates/jose/src/claims.rs
```diff
@@ -235,6 +239,82 @@ impl From<&TimeOptions> for TimeNotBefore {
...
+        | JsonWebSignatureAlg::Es512
+        | JsonWebSignatureAlg::Ps512 => {
+            let mut hasher = Sha512::new();
+            hasher.update(token);
+            let hash = hasher.finalize();
+            // Left-most half
+            hash.get(..32).map(ToOwned::to_owned)
+        }
+        JsonWebSignatureAlg::EdDsa => return Err(anyhow::anyhow!("unsupported algorithm")),
+        JsonWebSignatureAlg::None => return Ok(None),
```
sandhose, line 283: I think I'd return an error here instead. You're throwing an error anyway later if it returns None.

crates/handlers/src/oauth2/token.rs
```diff
@@ -343,16 +331,20 @@ async fn authorization_code_grant(
...
         let key = key_store
             .signing_key_for_algorithm(alg)
             .context("no suitable key found")?;
 
+        if let Some(hash) = hash_token(alg, &access_token_str)? {
+            claims::AT_HASH.insert(&mut claims, hash)?;
+        }
+        if let Some(hash) = hash_token(alg, &grant.code)? {
+            claims::C_HASH.insert(&mut claims, hash)?;
+        }
```
sandhose, line 346: Here it will never try to sign with the `one` alg: the key store will throw an error because no key will match the `none` alg

sandhose (changes requested):
I *think* it does not make sense to make `hash_token` return a `Result<Option<String>>` because of the case of the `none` alg, since it should never happen

crates/handlers/src/oauth2/token.rs
zecakeh: Right, because the `none` alg is not supported by the server.

crates/jose/Cargo.toml
```diff
@@ -9,6 +9,7 @@ license = "Apache-2.0"
 anyhow = "1.0.64"
 base64ct = { version = "1.5.2", features = ["std"] }
 chrono = { version = "0.4.22", features = ["serde"] }
+data-encoding = "2.3.2"
```
sandhose, line 12: I started using `base64ct` for base64 encoding instead of `data-encoding`, since it's used by RustCrypto crates, like `password-hash` and `pem-rfc7468`. I'll do a PR later to remove `data-encoding` completely, but if here you could already use `base64ct` instead that would be great

crates/jose/src/claims.rs
```diff
@@ -235,6 +239,81 @@ impl From<&TimeOptions> for TimeNotBefore {
...
+        }
+        JsonWebSignatureAlg::Hs384
+        | JsonWebSignatureAlg::Rs384
+        | JsonWebSignatureAlg::Es384
+        | JsonWebSignatureAlg::Ps384 => {
+            let mut hasher = Sha384::new();
+            hasher.update(token);
+            let hash = hasher.finalize();
+            // Left-most half
+            hash.get(..24).map(ToOwned::to_owned)
```
sandhose, line 272:
```suggestion
            hash.get(..24).cloned()
```

zecakeh: I tried this and it doesn't work.

sandhose: Nevermind, then :)

sandhose (approved):
Awesome, thanks!

PR #441, opened 2022-09-14
Allow access token types in introspection auth methods

zecakeh:
According to [RFC8414](https://www.rfc-editor.org/rfc/rfc8414#page-7).

### DIFF
--- a/crates/oauth2-types/src/oidc.rs
+++ b/crates/oauth2-types/src/oidc.rs
@@ -17,7 +17,7 @@ use std::ops::Deref;
 use language_tags::LanguageTag;
 use mas_iana::{
     jose::{JsonWebEncryptionAlg, JsonWebEncryptionEnc, JsonWebSignatureAlg},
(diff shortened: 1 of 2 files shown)


zecakeh:
I'm wondering if this is gonna work with #431...

crates/oauth2-types/src/oidc.rs
```diff
@@ -433,10 +483,18 @@ impl ProviderMetadata {
...
         }
 
+        let introspection_methods = metadata
+            .introspection_endpoint_auth_methods_supported
+            .as_ref()
+            .map(|v| {
+                v.iter()
+                    .filter_map(AuthenticationMethodOrAccessTokenType::authentication_method)
+                    .collect::<Vec<_>>()
+            });
```
sandhose, line 493: So basically here you're doing that to make it a Vec of only the auth methods, right? Could you just add a comment saying that the introspection methods can also include token types, and that's why you have to do that transformation first?

sandhose (changes requested):
If you could add a comment in the provider metadata validation that would be great, otherwise LGTM

PR #483, opened 2022-10-17
Fix imports used by feature-gated code

zecakeh:
Avoids cargo check errors and warning when the features are not enabled.


sandhose:
Whoops, thanks :)

PR #486, opened 2022-10-17
Fix RSA JWT signature and add snapshot tests for JWT signature

sandhose:
This upgrades the RSA crate to 0.7.0 (from 0.7.0-rc0) and use a released version of the sha2 crate.

It fixes the RS* and PS* JWT signature, and introduces snapshots for tests.
This helps verifying with other tools (like jwt.io) that the JWT we're signing are valid.


crates/jose/src/jwk/private_parameters.rs
```diff
@@ -271,7 +271,7 @@ mod rsa_impls {
         type Error = rsa::errors::Error;
         fn try_from(value: &RsaPrivateParameters) -> Result<Self, Self::Error> {
             let key: RsaPrivateKey = value.try_into()?;
-            Ok(Self::new(key))
+            Ok(Self::new_with_salt_len(key, <H as Digest>::output_size()))
```
sandhose, line 274: That was the main issue with PSS signatures. For RS* signatures, it might just be the difference between RSA 0.7.0 RC0 and RC1.

crates/jose/src/jwk/public_parameters.rs
```diff
@@ -283,102 +232,58 @@ mod rsa_impls {
 
 mod ec_impls {
```
sandhose, line 233: I removed some From impls to avoid buggy conversions, especially with RSA signatures

zecakeh:
It looks good.

Maybe we should add tests to check that we can generate a key, then sign a JWT with it, and then the signature verification works, since that was how the issue was found in the first place.

sandhose:
> Maybe we should add tests to check that we can generate a key, then sign a JWT with it, and then the signature verification works, since that was how the issue was found in the first place.

We already have tests, but in two different context: one where we sign and verify from the keystore, and the other where we sign and verify from a JWKS.
Problem was when you tested `mas-oidc-client` against the server, you were effectively signing with the keystore, and verifying with the JWKS.

I guess we could have a test where we sign from the keystore, and verify from the keystore JWKS.

zecakeh:
That's probably good because that's what happens in production, since I copied the code from `mas-handler` to generate the ID token.

PR #615, opened 2022-12-01
Use associated error type in claims validator instead of anyhow.

sandhose:


crates/jose/src/claims.rs
```diff
@@ -33,16 +32,19 @@ pub enum ClaimError {
     ValidationError {
         claim: &'static str,
         #[source]
-        source: anyhow::Error,
+        source: Box<dyn std::error::Error + Send + Sync + 'static>,
```
sandhose, line 35: I originally made that a generic type param, but that was too much of a headache to use

crates/jose/src/claims.rs
```diff
@@ -338,15 +366,20 @@ impl<'a, T> Contains<'a, T> {
     }
 }
 
+#[derive(Debug, Clone, Copy, Error)]
+#[error("OneOrMany doesn't contains value")]
```
zecakeh, line 370:
```suggestion
#[error("OneOrMany doesn't contain value")]
```

PR #616, opened 2022-12-02
Make the OIDC issuer a string instead of a URL

sandhose:
`url::Url` normalizes the URL a bit, by, for example, adding a trailing slash if the URL path is empty.
This is an issue for example with `https://account.google.com`, which doesn't have a trailing slash ; but then when verifying the `iss` in the ID token, it mismatches because of that normalization.

This changes the issuer in the provider metadata with a string instead.

cc @zecakeh


zecakeh (approved):
Makes sense!

PR #698, opened 2022-12-18
Don't enforce RS256 ID Token signing alg in provider metadata

zecakeh:
The [OIDC Discovery spec](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata) says that `id_token_signing_alg_values_supported` MUST contain the RS256 algorithm, which was the reason for this validation.

It made mas-oidc-client fail one of the tests of the OIDC conformance suite where the provider metadata would only include the ES256 algorithm. After [reporting it as a bug](https://gitlab.com/openid/conformance-suite/-/issues/1113), it was actually deemed an "overly zealous" validation, as it could be triggered by some servers, like the ones implementing FAPI, where the RS256 algorithm is not allowed.

### DIFF
--- a/crates/oauth2-types/src/oidc.rs
+++ b/crates/oauth2-types/src/oidc.rs
@@ -362,7 +362,7 @@ pub struct ProviderMetadata {
     /// JSON array containing a list of the JWS signing algorithms (`alg`
     /// values) supported by the OP for the ID Token.
     ///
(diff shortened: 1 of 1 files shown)


zecakeh:
ping @sandhose for review

sandhose (approved):

PR #760, opened 2023-01-09
Implement `Clone` on `ClientCredentials`

zecakeh:
Allows to reuse it for several requests.

Another option would be to take `&self` in `apply_to_request`.

### DIFF
--- a/crates/oidc-client/src/types/client_credentials.rs
+++ b/crates/oidc-client/src/types/client_credentials.rs
@@ -14,7 +14,7 @@
 
 //! Types and methods for client credentials.
 
-use std::{collections::HashMap, fmt};
+use std::{collections::HashMap, fmt, sync::Arc};
 
 use base64ct::{Base64UrlUnpadded, Encoding};
 use chrono::{DateTime, Duration, Utc};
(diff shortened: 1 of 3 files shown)


crates/oidc-client/src/types/client_credentials.rs
```diff
@@ -52,17 +52,17 @@ pub const CLIENT_SUPPORTED_AUTH_METHODS: &[OAuthClientAuthenticationMethod] = &[
 
 /// A function that takes a map of claims and a signing algorithm and returns a
 /// signed JWT.
-pub type JwtSigningFn =
-    dyn Fn(HashMap<String, Value>, JsonWebSignatureAlg) -> Result<String, BoxError> + Send + Sync;
+pub type JwtSigningFn = fn(HashMap<String, Value>, JsonWebSignatureAlg) -> Result<String, BoxError>;
```
sandhose, line 55: Doesn't that limit to only static functions, and not closures which capture variables from their environment? Wouldn't that make it harder to use in most cases?

zecakeh: Ah right I didn't read the fine prints :sweat_smile:

I don't know if using an `Arc` is too much though. I assume it's either this or making `apply_to_request` use a reference.

sandhose: The `Arc` is fine with me if it works 👍

sandhose (approved):

PR #1041, opened 2023-03-22
Allow redirect URIs with any port for loopback interfaces

zecakeh:
Part of #275.

### DIFF
--- a/crates/data-model/src/oauth2/client.rs
+++ b/crates/data-model/src/oauth2/client.rs
@@ -118,7 +118,7 @@ impl Client {
             ([], _) => Err(InvalidRedirectUriError::NoneRegistered),
             ([one], None) => Ok(one),
             (_, None) => Err(InvalidRedirectUriError::MultipleRegistered),
-            (uris, Some(uri)) if uris.contains(uri) => Ok(uri),
+            (uris, Some(uri)) if uri_matches_one_of(uri, uris) => Ok(uri),
             _ => Err(InvalidRedirectUriError::NotAllowed),
         }
     }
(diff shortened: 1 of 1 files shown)


crates/data-model/src/oauth2/client.rs
```diff
@@ -175,3 +175,66 @@ impl Client {
...
+            Url::parse("https://example.org").unwrap(),
+        ];
+
+        // Non-loopback interface URIs.
+        assert!(uri_matches_one_of(
+            &Url::parse("https://example.org").unwrap(),
+            registered_uris
+        ));
+        assert!(!uri_matches_one_of(
+            &Url::parse("https://example.com").unwrap(),
```
sandhose, line 222: It would probably be better to test here changing the port only?

crates/data-model/src/oauth2/client.rs
```diff
@@ -175,3 +175,66 @@ impl Client {
...
+                && registered_uri.path() == uri.path()
+            {
+                return true;
+            }
+        }
+
+        false
+    } else {
+        registered_uris.contains(uri)
+    }
```
sandhose, line 200: I'm tempted to just try to strip the port on the URI if it's localhost? That would require that the registered URI don't have a port in them though... which might be fine? I can't really find a good source in RFCs that mixes dynamic registration and native apps with localhost/127.0.0.1 as the redirector :(

Something like that?
```suggestion
    if LOCAL_HOSTS.contains(&uri.host_str().unwrap_or_default()) {
        let mut uri = uri.clone();
        // Try matching without the port first
        if url.set_port(None).is_ok() {
            if registered_uris.contains(uri) {
                return true;
            }
        }
    } 
    
    registered_uris.contains(uri)
```

sandhose (approved):
Some nitpicks but overall LGTM

crates/data-model/src/oauth2/client.rs
zecakeh: Yes, the clearest I got was from [OAuth 2.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-08#name-registration-requirements):

> Authorization servers MUST reject authorization requests that specify a redirect URI that doesn't exactly match one that was registered, with an exception for loopback redirects, where an exact match is required except for the port URI component.

Which to me means we should check everything but the port. But if you prefer it your way, it's fine by me.

PR #1173, opened 2023-05-09
Make ClientErrorCode.error_description optional.

pixlwave:
It's an optional parameter and deserialization of an authorization error response was failing due to this.

https://www.rfc-editor.org/rfc/rfc6749#section-4.1.2.1

### DIFF
--- a/crates/oauth2-types/src/errors.rs
+++ b/crates/oauth2-types/src/errors.rs
@@ -30,7 +30,8 @@ pub struct ClientError {
     pub error: ClientErrorCode,
 
     /// A human-readable description of the error.
-    pub error_description: Cow<'static, str>,
+    #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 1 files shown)


crates/oauth2-types/src/errors.rs
```diff
@@ -30,7 +30,7 @@ pub struct ClientError {
     pub error: ClientErrorCode,
 
     /// A human-readable description of the error.
-    pub error_description: Cow<'static, str>,
+    pub error_description: Option<Cow<'static, str>>,
```
zecakeh, line 33: This removes the field instead of having it set as `null` when serializing:

```suggestion
    #[serde(skip_serializing_if = "Option::is_none")]
    pub error_description: Option<Cow<'static, str>>,
```

pixlwave: Ah yeah, good point I didn't think about serialising. Updated.

PR #1178, opened 2023-05-10
Downgrade tracing to 0.1.37

zecakeh:
Version 0.1.38 was yanked and no new version was released. So that makes the crates impossible to compile if the dependency was not cached!

Apparently it included an unwanted breaking change: https://github.com/tokio-rs/tracing/issues/2578.


sandhose (approved):
Thanks!

Issue #1375, opened 2023-07-19
Feature request: Simplify the setup process


dklimpel:
I have a few comments that I as an server admin notice and can possibly be improved to make it easier for the community to use the module in the future.

I will probably update the ticket in the future.

- [x] currently no tagged versions
- [ ] delivery of built NodeJS files as artifact
- [ ] delivery of built default OPA policy file (policy.wasm)
- [x] prebuild rust binaries similar to Synapse?

Differences in documentation
> Run the server via cargo run -- server -c config.yaml
https://github.com/matrix-org/matrix-authentication-service#running

vs.

> mas-cli server
https://matrix-org.github.io/matrix-authentication-service/usage/usage.html#running

The [official documentation](https://matrix-org.github.io/matrix-authentication-service/) has no infromation about:
- NodeJS
- OPA

sandhose:
The documentation and distribution is definitely lacking. Although it is already in my todo-list to make that better in the next few weeks, contributions on that front is definitely appreciated! Feel free to ask on [#matrix-auth:matrix.org](https://matrix.to/#/#matrix-auth:matrix.org) if something needs to be clarified.

At some point, we were embedding all the assets in the binary, which did not scale well. 

For the record, for ~production, I'm currently relying on the nightly docker images. The Docker image is a good source to know what's needed to run it, outside of the binary itself:

 - `/usr/local/share/mas-cli/assets/` has the built frontend assets
 - `/usr/local/share/mas-cli/manifest.json` is the Vite manifest generated for the frontend
 - `/usr/local/share/mas-cli/templates/` has the `templates/` folder copied as-is
 - `/usr/local/share/mas-cli/policy.wasm` has the OPA policy compiled to WASM

I'm definitely down to publish full archives with both the binary and all the assets.

About publishing versions, I'm not sure when the service will be stable enough to cut a clear version with proper changelogs between versions.

A few questions we would need to answer:

 - which platform+arch are we distributing binaries for? Synapse currently does Linux and macOS, on x86_64, i686 and arm64. I think would skip i686.
 - on Linux, do we build for `glibc` (`{aarch64,x86_64}-unknown-linux-gnu`) based binaries, or do we build fully-static binaries based on `musl` (`{aarch64,x86_64}-unknown-linux-musl`)? Note that in the Docker image we're using `musl` binaries.
 - should we rely on system certificates for root CAs or embed them from webpki in the binary?
[...]

PR #1379, opened 2023-07-20
update installation documentation

dklimpel:
Related to:
- #1375

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,31 +7,14 @@ See the [Documentation](https://matrix-org.github.io/matrix-authentication-servi
 
 ## Running
 
-- [Install Rust and Cargo](https://www.rust-lang.org/learn/get-started)
-- [Install Node.js and npm](https://nodejs.org/)
-- [Install Open Policy Agent](https://www.openpolicyagent.org/docs/latest/#1-download-opa)
-- Clone this repository
-- Build the frontend
-  ```sh
-  cd frontend
-  npm ci
-  npm run build
-  cd ..
-  ```
-- Build the Open Policy Agent policies
-  ```sh
-  cd policies
-  make
-  # OR, if you don't have `opa` installed and want to build through the OPA docker image
-  make DOCKER=1
-  cd ..
-  ```
-- Generate the sample config via `cargo run -- config generate > config.yaml`
-- Run a PostgreSQL database locally
-  ```sh
-  docker run -p 5432:5432 -e 'POSTGRES_USER=postgres' -e 'POSTGRES_PASSWORD=postgres' -e 'POSTGRES_DATABASE=postgres' postgres
-  ```
-- Update the database URI in `config.yaml` to `postgresql://postgres:postgres@localhost/postgres`
-- Run the database migrations via `cargo run -- database migrate`
-- Run the server via `cargo run -- server -c config.yaml`
+- [Observe and install requirements](https://matrix-org.github.io/matrix-authentication-service/usage/installation.html#requirements)
+- [Install Matrix Authentication Service](https://matrix-org.github.io/matrix-authentication-service/usage/installation.html#installing-from-the-source)
+- [Generate the sample config](https://matrix-org.github.io/matrix-authentication-service/usage/configuration.html)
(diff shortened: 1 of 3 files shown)


docs/usage/configuration.md
```diff
@@ -97,14 +102,32 @@ Controls the web server.
 
 ```yaml
 http:
-  # On what address and port the server should listen to
-  address: 0.0.0.0:8080
-
-  # Path from which to serve static files
-  web_root: /var/www/static
-
+  listeners:
```
sandhose, line 105: This section could use more comments, but I'll leave it like that for now

docs/usage/configuration.md
```diff
@@ -32,6 +32,11 @@ docker run --rm -v `pwd`/config.yaml:/config.yaml \
 
 Note that with Docker, the config file must be mounted inside the container with `-v/--volume`.
 
+Update the database URI in `config.yaml` to your database,
+e.g. `postgresql://postgres:postgres@localhost/postgres`
+
+See also the next paragraph for a [minimum configuration ](#minimal-configuration)
```
sandhose, line 38:
```suggestion
See also the next section for a [minimal configuration](#minimal-configuration)
```

docs/usage/installation.md
```diff
@@ -2,10 +2,12 @@
 
 ## Requirements
```
sandhose, line 3: I think we should distinguish build deployments (NodeJS, OPA, a Rust toolchain) from runtime ones (just a Postgres database), I'll try to clarify that later

README.md
```diff
@@ -7,31 +7,14 @@ See the [Documentation](https://matrix-org.github.io/matrix-authentication-servi
...
+- [Install Matrix Authentication Service](https://matrix-org.github.io/matrix-authentication-service/usage/installation.html#installing-from-the-source)
+- [Generate the sample config](https://matrix-org.github.io/matrix-authentication-service/usage/configuration.html)
+- [Provide the database](https://matrix-org.github.io/matrix-authentication-service/usage/installation.html#database)
+- [Customize your minimal configuration](https://matrix-org.github.io/matrix-authentication-service/usage/configuration.html#minimal-configuration)
+  - `database.uri`
+- [Run the database migrations](https://matrix-org.github.io/matrix-authentication-service/usage/usage.html#running)
+- [Run the server](https://matrix-org.github.io/matrix-authentication-service/usage/usage.html#running)
 - Go to <http://localhost:8080/>
+
+- Or use the [docker image](https://matrix-org.github.io/matrix-authentication-service/usage/installation.html#running-from-the-docker-image) alternatively.
```
sandhose, line 20: This whole section is very weird to me. I think this should stay a quick "how to get up and running" for deployment, and have dedicated documentation on how to run in production. Can you revert that change for now?

sandhose (changes requested):
Thanks for the contribution! There is a lot of work to get the documentation in the right place, and this is a good first step toward that.

README.md
dklimpel: I will revert it.

My concern was that there are two instructions in different places and the content is not the same and also both need to be maintained.

docs/usage/installation.md
```diff
@@ -60,3 +81,44 @@ Note that when running in a Docker environment
...
+sudo -u postgres bash
+```
+
+Then, create a postgres user and a database with:
+```
+# this will prompt for a password for the new user
+createuser --pwprompt matrix_authentication_user
+
+createdb --owner=matrix_authentication_user matrix_authentication
+```
```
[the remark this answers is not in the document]
dklimpel: My understanding was that this deployment is a prerequisite for running Synapse in the very near future. Not every Synapse provider uses kubernetes or docker. And if the transition is to be easy for the community, it should be clear that this database is needed and how to create it. This documentation is also part of Synapse.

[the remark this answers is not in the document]
sandhose: Alright, it felt like those instructions could be distro-specific, but seeing that Synapse has similar documentation, I'm happy with it. We can probably expand on it later anyway

sandhose (approved):

PR #1454, opened 2023-08-05
Add type support for Matrix CS API guest scope

zecakeh:
According to [MSC2967](https://github.com/matrix-org/matrix-spec-proposals/pull/2967)


sandhose (changes requested):
Please add some tests for this one, especially since the parsing/serialising of all this is not that obvious

Otherwise LGTM

sandhose (approved):

PR #1456, opened 2023-08-05
Add client-side support for software statement, ID and version

zecakeh:
According to [RFC7591](https://datatracker.ietf.org/doc/html/rfc7591).


crates/oidc-client/src/requests/registration.rs
```diff
@@ -46,11 +59,23 @@ pub async fn register_client(
     http_service: &HttpService,
     registration_endpoint: &Url,
     client_metadata: VerifiedClientMetadata,
+    software_statement: Option<String>,
```
sandhose, line 62: Why is it not part of the VerifiedClientMetadata directly?

sandhose (changes requested):
I'm fine with the `software_{id,version}`, but I'm not sure I understand why the `software_statement` is not part of the client metadata directly and handled somewhere else?

crates/oidc-client/src/requests/registration.rs
zecakeh: That's because in my head, you could use `VerifiedClientMetadata` is also anything you could put as a claim in the software statement, so you can't put this field in it. But maybe that doesn't make sense :sweat_smile:.

sandhose: Yeah, makes sense. I'll probably touch that again once we actually start to care about the software_statement in MAS 👍

sandhose (approved):

Issue #20, opened 2022-01-17
Tool to import Synapse user DB


sandhose:
At some point we should import the existing users from Synapse's database. It means we need to support Synapse's existing password hashing scheme

hughns:
syn2mas now exists so let's close this issue. If there enhancements needed then we can make new issues.

PR #1512, opened 2023-08-11
syn2mas tool to import users from Synapse

hughns:
Implements https://github.com/matrix-org/matrix-authentication-service/issues/20


hughns:
@sandhose I don't know how you want the docker bake set up to publish two images so that isn't included in the build.yaml.

PR #1526, opened 2023-08-15
Introduce Base64 type

jplatte:
There's two things I'm uncertain about:

- Does it make sense to use the base64ct crate? Presumably it's slower because it uses a constant-time algorithm. Could also use the base64 crate like serde_with does, it would be another direct dependency but lots of other crates in the dependency graph depend on it anyways.
- Do the changes to the constructor functions I did make sense? It could be annoying for callers outside this crate to have to wrap a byte vector in `Base64UrlNoPad::new()` before calling these, though inside this crate it was convenient.


crates/jose/Cargo.toml
```diff
@@ -22,7 +22,7 @@ schemars = "0.8.12"
 sec1 = "0.7.3"
 serde.workspace = true
 serde_json.workspace = true
-serde_with = { version = "3.2.0", features = ["base64"] }
+serde_with = { version = "3.2.0" }
```
sandhose, line 25:
```suggestion
serde_with = "3.2.0"
```

for consistency

crates/jose/src/jwk/public_parameters.rs
```diff
@@ -113,27 +105,24 @@ impl ParametersInfo for RsaPublicParameters {
...
 
     #[schemars(with = "String")]
-    #[serde_as(as = "Base64<UrlSafe, Unpadded>")]
-    y: Vec<u8>,
+    y: Base64UrlNoPad,
 }
 
 impl EcPublicParameters {
-    pub const fn new(crv: JsonWebKeyEcEllipticCurve, x: Vec<u8>, y: Vec<u8>) -> Self {
+    pub const fn new(crv: JsonWebKeyEcEllipticCurve, x: Base64UrlNoPad, y: Base64UrlNoPad) -> Self {
```
sandhose, line 125: I think I would prefer to keep the bytes parameters as `Vec<u8>` since the base64 is only a serialization/deserialization detail? It does mean that `Base64::new` should be `const` though.

This is not a strong opinion, if it really was more convenient to have them like that, it's fine with me

sandhose (approved):
Thanks!

I'm fine with using the `base64ct` crate, this isn't really a hot path in terms of performance anyway.

PR #1558, opened 2023-08-21
Note that pgcrypto is required

### LINKED ISSUE #1557: Docs don't mention that pgcrypto extension is required
```
$ docker run --network=host -v $(pwd)/tests-e2e/mas-config.yaml:/config.yaml ghcr.io/matrix-org/matrix-authentication-service:main database migrate

Error: could not run migrations

Caused by:
    0: while executing migrations: error returned from database: extension "pgcrypto" is not available
    1: error returned from database: extension "pgcrypto" is not available
    2: extension "pgcrypto" is not available
```

Looking at https://matrix-org.github.io/matrix-authentication-service/setup/database.html

DMRobertson:
Fixes #1557.

### DIFF
--- a/docs/setup/database.md
+++ b/docs/setup/database.md
@@ -28,6 +28,13 @@ createdb --owner=mas_user mas
 
 The above will create a user called `mas_user` with a password of your choice, and a database called `mas` owned by the `mas_user` user.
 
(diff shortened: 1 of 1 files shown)


sandhose:
As of #1585, pgcrypto shouldn't be required anymore

PR #1567, opened 2023-08-23
Rename 'Emails' route to 'Profile'

kerryarchibald:
Fixes https://github.com/matrix-org/matrix-authentication-service/issues/1547

- `/account/emails` -> `/account/profile`
- rename 'Emails' nav item to 'Profile'

Screenshot 2023-08-23 at 17 07 18


frontend/src/Router.tsx
```diff
@@ -99,8 +99,12 @@ const segmentsToRoute = (segments: string[]): Route => {
     return { type: "home" };
   }
 
+  // legacy support for /emails
```
sandhose, line 102: I'm not sure this is needed tbh

frontend/src/Router.tsx
```diff
@@ -183,8 +187,9 @@ const InnerRouter: React.FC = () => {
   switch (route.type) {
     case "home":
       return <Home />;
-    case "email-list":
-      return <EmailList />;
+    // legacy
```
sandhose, line 190: How is that `legacy`?

sandhose (approved):
I don't it's worth keeping the match on `/emails` but other than that LGTM

Issue #1631, opened 2023-08-29
Cannot distinguish inactive browser sessions


kerryarchibald:
When viewing browser sessions without 'active only' filter, there is no way to distinguish active and inactive sessions.

AC:
- `BrowserSession` includes `finishedAt` property.
- Browser sessions tiles in the UI include `finished at` metadata with relative time

PR #1641, opened 2023-08-29
Fix: cannot distinguish inactive Browser sessions

kerryarchibald:
Fixes https://github.com/matrix-org/matrix-authentication-service/issues/1631

Screenshot 2023-08-29 at 20 42 25


sandhose (approved):
LGTM!

Issue #1639, opened 2023-08-29
Switch profile and sessions tabs


Johennes:
The profile tab should be the first / default one, with the sessions tab second.

PR #1653, opened 2023-08-30
Make Profile page the default route

kerryarchibald:
Fixes https://github.com/matrix-org/matrix-authentication-service/issues/1639

Screenshot 2023-08-30 at 16 41 51

- Reorders tabs
- Makes default route (`/account`) show the profile page
- Makes sessions route `/account/sessions-overview`


sandhose (approved):
Would you mind just renaming the `<Home />` component/page to something that makes more sense? Then I'm fine if you merge it yourself :)

PR #1673, opened 2023-09-01
Implement MSC2965 action parameter

kerryarchibald:
Fixes https://github.com/matrix-org/matrix-authentication-service/issues/1550

Redirects to session detail when query param `action=session_end` is encountered.

https://github.com/matrix-org/matrix-authentication-service/assets/3055605/e1ec182e-7310-4127-8434-bad217811b19


frontend/src/Router.tsx
```diff
@@ -169,14 +143,32 @@ export const routeAtom = atom(
...
+ * Get route from path
+ * @returns Route
+ */
+const useRouteWithRedirect = (): Route => {
+  const location = useAtomValue(locationAtom);
+  const redirect = getRouteActionRedirection(location);
+
+  const [route, setRoute] = useAtom(routeAtom);
+  if (redirect) {
+    setRoute(redirect.route, redirect.searchParams);
```
sandhose, line 166: I think this should be in a useEffect, because you're not supposed to do state changes in the render?

Johennes: Kerry won't see this before Monday, sadly, and we have to land this PR today. @t3chguy could you help us out by giving advice?

sandhose: I can take care of finishing that up, alongside the server side of this

t3chguy: Yes I believe it should be in a useEffect

sandhose:
Alright, I: 
 - refactored a bit the router
 - did the backend part of the redirect so that params are kept through the login
 - implemented the other `action` defined in the MSC on the frontend

PR #1674, opened 2023-09-01
Standardise consent screens

### LINKED ISSUE #1638: Improve consent screen for compatibility sessions
The current screen is quite bare bones. We should:

* Update it to be more similar to the one for OAuth sessions
* Remove Matrix branding
* Display the client URI instead of the redirect URI

hughns:
Fixes #1638

I made the consent descriptions reusable

image
image


pmaier1 (approved):
:+1:

sandhose (approved):

PR #1677, opened 2023-09-01
Compoundify the templates

sandhose:
I also noticed the react app and the rest were sharing the same CSS file. This is becase CSS code split is broken, so I made sure that we're not lazy-loading pages anymore until this is fixed.


hughns (approved):
Looks great.

PR #1685, opened 2023-09-04
Parse browser session `userAgent` for session name

kerryarchibald:
Fixes https://github.com/matrix-org/matrix-authentication-service/issues/1635

Screenshot 2023-09-06 at 08 25 28

- adds user agent parsing (copied from EW)
- Use parsed UA for browser session title `<client> on <device> <deviceOS>`
- Add pill-style for 'current' label on current browser session


frontend/src/components/BrowserSession.tsx
```diff
@@ -90,7 +96,24 @@ const BrowserSession: React.FC<Props> = ({ session, isCurrent }) => {
     });
   };
 
-  const sessionName = isCurrent ? "Current browser session" : "Browser session";
+  const deviceInformation = parseUserAgent(data.userAgent || undefined);
+  const name =
+    sessionNameFromDeviceInformation(deviceInformation) || "Browser session";
+  const sessionName = isCurrent ? (
```
sandhose, line 102: Can you instead make a `current` prop on the `Session` component, so that it can be displayed on top of the session name and not inline?

sandhose (changes requested):
I love this, this is so much better!

Maybe a comment from @Johennes or @pmaier1 on the wording/info to show? It feels like the browser version and sometimes the device vendor are not always relevant

frontend/src/components/BrowserSession.tsx
kerryarchibald: Screenshot 2023-09-05 at 12 33 20

sandhose (approved):
Looks good, thanks!

PR #1708, opened 2023-09-06
Compoundify the consent screens

sandhose:
## For a legacy client

image

## For Element X iOS

image

## For Synapse Admin

image


hughns (approved):

hughns:
A great step forward and we can refine further from here.

sandhose:
@hughns and @pmaier1 said it looks great, we can fix the small stuff (the profile icon being ugly & the send icon not being available) in a subsequent PR

PR #1723, opened 2023-09-08
Display client logo

kerryarchibald:
Fixes https://github.com/matrix-org/matrix-authentication-service/issues/1672

Screenshot 2023-09-08 at 13 10 13

### DIFF
--- a/frontend/src/components/Session/ClientAvatar.module.css
+++ b/frontend/src/components/Session/ClientAvatar.module.css
@@ -13,28 +13,11 @@
  * limitations under the License.
  */
 
-.list {
-    display: flex;
-    flex-direction: column;
-    margin-top: var(--cpd-space-1x);
-    gap: var(--cpd-space-1x);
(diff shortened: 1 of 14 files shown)


frontend/src/components/Session/ClientAvatar.module.css
```diff
@@ -0,0 +1,23 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.avatar {
+    object-fit: cover;
+    overflow: hidden;
+    aspect-ratio: 1 / 1;
+    width: var(--mas-avatar-size);
+    border-radius: 50%;
```
sandhose, line 21: I'm not sure about that. Works well for our logos, but in other places (like the consent screen), we're not making logo rounds. I'll leave it like this for now, and we'll probably iterate on this later

sandhose (approved):

PR #1732, opened 2023-09-11
DM: Browser session detail page

kerryarchibald:
Screenshot 2023-09-11 at 16 37 14


frontend/src/utils/session/useCurrentBrowserSessionId.ts
```diff
@@ -0,0 +1,46 @@
...
+import { useAtomValue } from "jotai";
+
+import { currentBrowserSessionIdAtom } from "../../atoms";
+import { isOk, unwrapOk, unwrapErr, isErr } from "../../result";
+
+/**
+ * Query the current browser session id
+ * and unwrap the result
+ */
+export const useCurrentBrowserSessionId = (): {
```
sandhose, line 25: I don't understand the value of this, especially looking at the usage? Is it for easier mocking? If that's the case, you can instead inject the atom value with [`useHydrateAtoms`](https://jotai.org/docs/utilities/ssr)

sandhose (approved):
Other than the `useCurrentBrowserSessionId` hook that I don't understand, looks good to me!

frontend/src/utils/session/useCurrentBrowserSessionId.ts
kerryarchibald: The idea is to avoid having to do the error checking/handling and unwrapping in components. After adding the [error boundary](https://github.com/matrix-org/matrix-authentication-service/pull/1743) I simplify this to throw on error, so consuming components don't have to check the error and render the error page. 
https://github.com/matrix-org/matrix-authentication-service/pull/1743/files#diff-0c5b8a4159d243e8dd602e3530f1e41e9e6ec98cc0514317fd7abda2350f5369

sandhose: Alright, makes sense

PR #1736, opened 2023-09-11
Don't show redirect_uri on consent screen if not HTTP

hughns:
Because we don't know if it is validated by the OS at all and so could be misleading to the user

### DIFF
--- a/templates/pages/consent.html
+++ b/templates/pages/consent.html
@@ -32,7 +32,7 @@ limitations under the License.
         {% endif %}
 
         <h1 class="cpd-text-primary cpd-text-heading-xl-semibold"><a target="_blank" href="{{ client.client_uri }}">{{ client_name }}</a></h1>
-        <p class="cpd-text-secondary cpd-text-body-lg-regular"><span class="whitespace-nowrap">at {{ grant.redirect_uri | simplify_url }}</span> wants to access your account. This will allow <span class="whitespace-nowrap">{{ client_name }}</span> to:</p>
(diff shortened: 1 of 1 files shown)


sandhose (changes requested):
For now it is the only bit of minuscule trust we can put on that screen, and I don't want to get rid of it, even though it is flawed.
I would suggest showing **at the very least** the `client_uri`, since: 

 - [the OIDC dynamic reg spec says](https://openid.net/specs/openid-connect-registration-1_0.html#ClientMetadata) about it:
   > If present, the server SHOULD display this URL to the End-User in a followable fashion.
 - [the RFC for native apps says](https://datatracker.ietf.org/doc/html/rfc8252#section-7.1) (and we're enforcing it):
   >  When choosing a URI scheme to associate with the app, apps MUST use a URI scheme based on a domain name under their control, expressed in reverse order

Moreover, EXI is working towards using universal links anyway, so for Element clients it becomes completely irrelevant

hughns:
> When choosing a URI scheme to associate with the app, apps MUST use a URI scheme based on a domain name under their control, expressed in reverse order

The problem is that nothing is validating that the given domain is under control of the app owner. The app can choose any domain and give a matching `client_uri` and reverse `redirect_uri`.

As such, displaying a reverse domain is of zero benefit.

What is the trust that showing it brings?


Where the value is validated I do see benefit in showing it which is why my PR shows it if it is an HTTP URI.

pmaier1 (approved):
👍  from product side

PR #1743, opened 2023-09-11
Error boundary

kerryarchibald:
Use an error boundary to capture thrown errors and render an error page.


frontend/src/components/CompatSessionList.tsx
```diff
@@ -116,7 +115,8 @@ const CompatSessionList: React.FC<{ userId: string }> = ({ userId }) => {
   const [prevPage, nextPage] = useAtomValue(paginationFamily(userId));
   const [filter, setFilter] = useAtom(filterAtom);
 
-  if (isErr(result)) return <GraphQLError error={unwrapErr(result)} />;
+  if (isErr(result)) throw unwrapErr(result);
```
sandhose, line 118: Maybe we could add a util like

```ts
const unwrap = <T, E>(result: Result<T, E>): T => {
  if (isErr(result)) {
    throw unwrapErr(result);
  }
  return unwrapOk(result);
}
```

And then this would become 

```suggestion
  const compatSessionList = unwrap(result);
```

sandhose (approved):

Issue #1750, opened 2023-09-12
Destructive actions in the frontend don't have any confirmation dialogs


sandhose:
This includes:
 - removing an email
 - ending all kinds of sessions

PR #1759, opened 2023-09-13
Confirm destructive actions

kerryarchibald:
Fixes https://github.com/matrix-org/matrix-authentication-service/issues/1750

Screenshot 2023-09-13 at 13 57 44
Screenshot 2023-09-13 at 13 44 04


sandhose (approved):
That is really nice!

How annoying would it be to base the implementation on [`@radix-ui/react-alert-dialog`](https://www.radix-ui.com/primitives/docs/components/alert-dialog) instead, since most of compound is basing their accessibility stuff on it?

I'm 100% OK with postponing using Radix to later if you don't want to do that now, so feel free to merge if that's the case

kerryarchibald:
@sandhose good thought re radix. Updated to use radix - the radix composition style is new to me and I'm not sure I like the inversion of control.

sandhose (approved):
Looks good, thanks!

PR #1764, opened 2023-09-13
Profile: hide email pagination controls when no other pages

kerryarchibald:
Screenshot 2023-09-13 at 15 50 59
Screenshot 2023-09-13 at 15 49 33


frontend/src/components/UserProfile/UserEmailList.tsx
```diff
@@ -178,6 +172,14 @@ const UserEmailList: React.FC<{
           onRemove={onRemove}
         />
       ))}
+      {(prevPage || nextPage) && (
```
kerryarchibald, line 175: I wonder if it would be good to still display a total when there is just one page?

sandhose (approved):
LGTM! Maybe we can hide on the component level, and if you think it does not make sense to do so, feel free to merge

PR #1773, opened 2023-09-14
Go back to session list from session detail

kerryarchibald:
Fixes https://github.com/matrix-org/matrix-authentication-service/issues/1760

Adds a 'back to sessions' button to all session detail screens.
This links back to session overview from all session types, as session lists by type will be removed soon with joint session list.

Screenshot 2023-09-14 at 14 25 58
Screenshot 2023-09-14 at 14 34 25


frontend/src/components/SessionDetail/SessionHeader.tsx
```diff
@@ -0,0 +1,36 @@
...
+import styles from "./SessionHeader.module.css";
+
+const SessionHeader: React.FC<PropsWithChildren> = ({ children }) => (
+  <header className={styles.header}>
+    <Link
+      kind="button"
+      route={{ type: "sessions-overview" }}
+      title="Back to sessions"
+    >
+      <IconChevron height="1.5em" width="1.5em" className={styles.backIcon} />
```
sandhose, line 30: There is an `<IconButton />` component in Compound, but with our current `<Link />` component it's annoying to use. Let's leave it like this for now

sandhose (approved):
LGTM, thanks!

frontend/src/components/SessionDetail/SessionHeader.tsx
sandhose: I ended up extracting the navigation logic in a hook

PR #1784, opened 2023-09-15
Session: Display IP address and last active timestamp

kerryarchibald:
For https://github.com/matrix-org/matrix-authentication-service/issues/1716

Screenshot 2023-09-15 at 16 08 49
Screenshot 2023-09-15 at 16 00 51

### DIFF
--- a/frontend/codegen.ts
+++ b/frontend/codegen.ts
@@ -21,6 +21,14 @@ const config: CodegenConfig = {
   generates: {
     "./src/gql/": {
       preset: "client",
+      config: {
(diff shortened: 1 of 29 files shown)


sandhose:
@kerryarchibald I hijacked your PR to wire up the last IP address and last active timestamp

I also cleaned up fragment usages, and configured the GraphQL codegen so that the `DateTime` scalars are typed as `string` instead of `any`. Unknown scalars will now be typed as `unknown` instead of `any`

PR #1785, opened 2023-09-15
Deny URIs from client that are public suffixes & allow subdomains

sandhose:
This does two things:

 - check that hostnames in URIs are not public suffixes, so no one can claim to be `co.uk`, `github.io` or `s3.amazonaws.com`
 - allow subdomains in client registrations, so you can claim to be `https://example.com/`, have your redirect URI be `https://app.example.com/callback` and your logo hosted on `https://assets.example.com/logo.png`

### DIFF
--- a/crates/handlers/src/oauth2/registration.rs
+++ b/crates/handlers/src/oauth2/registration.rs
@@ -25,9 +25,11 @@ use oauth2_types::{
         ClientMetadata, ClientMetadataVerificationError, ClientRegistrationResponse, Localized,
     },
 };
+use psl::Psl;
 use rand::distributions::{Alphanumeric, DistString};
 use thiserror::Error;
 use tracing::info;
+use url::Url;
 
 use crate::impl_from_error_for_route;
 
@@ -41,6 +43,9 @@ pub(crate) enum RouteError {
 
     #[error("invalid client metadata")]
     InvalidClientMetadata(#[from] ClientMetadataVerificationError),
+
+    #[error("{0} is a public suffix, not a valid domain")]
+    UrlIsPublicSuffix(&'static str),
 
     #[error("denied by the policy: {0:?}")]
     PolicyDenied(Vec<Violation>),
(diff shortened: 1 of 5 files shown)


crates/handlers/src/oauth2/registration.rs
```diff
@@ -219,8 +310,28 @@ mod tests {
...
+        }
+
+        assert!(url_is_public_suffix("https://.com."));
+        assert!(url_is_public_suffix("https://co.uk"));
+        assert!(url_is_public_suffix("https://github.io"));
+        assert!(!url_is_public_suffix("https://example.com"));
+        assert!(!url_is_public_suffix("https://matrix-org.github.io"));
+        assert!(!url_is_public_suffix("http://localhost"));
+        assert!(!url_is_public_suffix("org.matrix:/callback"));
+        assert!(!url_is_public_suffix("http://somerandominternaldomain"));
```
sandhose, line 333: That's one relevant test to understand what's considered to be a public suffix and what is not

clokep: I'm not convinced the `host.len()` check is tested anywhere?

sandhose: What do you mean? The empty host is tested by `org.matrix:/callback`, and that length test is there to cover the `http://.com/` case: in that case, the prefix will be `com` but the host will be `.com`. If we were doing `host.len() != suffix.len()`, we would miss that case.
Effectively, that inequality is a shorter version `(host.len() != suffix.len()) && (host.len() != (suffix.len() + 1))`

clokep: Ah, I misunderstood what that's checking. I can see how that's being hit now.

sandhose: I made the comment around that logic more obvious in https://github.com/matrix-org/matrix-authentication-service/pull/1785/commits/7141d8692245dd31de795fe3c843e5cae79f0525

PR #1794, opened 2023-09-18
frontend: upgrade compound & simplify the confirmation modal implementation

sandhose:
Now that refs are correctly forwarded in Compound (vector-im/compound-web#76), we don't have to do a hack to have the autofocus working anymore


frontend/src/components/ConfirmationModal/ConfirmationModal.tsx
```diff
@@ -40,7 +40,7 @@ type Props = {
 /**
  * Generic confirmation modal
  * controls its own open state
- * calls onDeny on cancel, esc, or overlay click
```
sandhose, line 43: Looks like Radix' `AlertDialog` is intentionally not dismissible from clicking on the overlay. They do have that behaviour on regular `Dialog`s.

kerryarchibald (approved):
Thank you!

PR #1938, opened 2023-10-10
syn2mas: use zod to validate the config files

sandhose:
This uses zod to actually validate the config files instead of assuming they are correct.


tools/syn2mas/src/advisor.mts
```diff
@@ -42,111 +60,205 @@ export async function advisor(argv?: string[]): Promise<void> {
...
   async function count(query: Knex.QueryBuilder): Promise<number> {
-    const res = await (query.first());
+    const res = await query.first();
     if (!res) {
       return 0;
     }
     return res["count(*)"] as number;
   }
 
-  const adminUsers = await count(synapse.count("*").from<SUser>("users").where({ admin: 1 }));
```
sandhose, line 58: I think I formatted with prettier without noticing…

tools/syn2mas/package.json
```diff
@@ -8,23 +8,27 @@
...
     "eslint-config-prettier": "^9.0.0",
-    "eslint-plugin-import": "^2.28.1",
+    "eslint-import-resolver-typescript": "^3.6.1",
+    "eslint-plugin-import": "npm:eslint-plugin-i@^2.28.1",
     "eslint-plugin-matrix-org": "^1.2.1",
+    "eslint-plugin-prettier": "^5.0.0",
     "eslint-plugin-unicorn": "^48.0.1",
     "prettier": "^3.0.3",
-    "ts-node": "^10.9.1",
+    "ts-node": "github:TypeStrong/ts-node#main",
```
sandhose, line 31: There is a bug in the npm released version which makes it crash when there is an array of `extends` in the tsconfig

hughns (approved):

PR #1962, opened 2023-10-17
Wire up i18n for the React frontend

t3chguy:


sandhose (approved):
Alright, I think I got to the bottom of this 👍 

I did a few small changes, mainly:

 - include the i18next-parser config in the lint/tsc pipeline, so that it gets formatted and typechecked
 - load the translations through the vite assets pipeline, not just the public folder, so that they get unique hashes for better caching

PR #1990, opened 2023-10-20
fix(syn2mas): use postgres host when connecting

cleverer:
For dockerized setups syn2mas is unusable, since the database most likely is not accessible on localhost. Therfore we also need to set the actual host parameter of the postgres config when connecting.


sandhose (approved):
You are absolutely right, thanks for catching this one!

PR #2047, opened 2023-11-03
jose: Reduce stack size of JsonWebSignatureHeader

jplatte:
… by putting the optional jwk field behind a box.

The overall size will be one pointer larger when the field is Some(_), but more than 300 bytes small when it is None.


sandhose (approved):
Makes sense, thanks!

PR #2048, opened 2023-11-03
syn2mas: support for deactivated users & use timestamps when generating IDs

### LINKED ISSUE #2032: syn2mas should be able to import deactivated users
Now that we can lock users, syn2mas should import deactivated users as locked ones.
It should skip importing any access tokens for those.

sandhose:
Fixes #2032

It adds support for deactivated users, and uses object creation timestamps when generating ULIDs

### DIFF
--- a/tools/syn2mas/src/migrate.mts
+++ b/tools/syn2mas/src/migrate.mts
@@ -119,9 +119,9 @@ export async function migrate(): Promise<void> {
     fatals += 1;
   }
 
-  function makeUuid<T>(): UUID<T> {
-    return id128.Uuid4.fromRaw(
-      id128.UlidMonotonic.generate().toRaw(),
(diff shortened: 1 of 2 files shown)


tools/syn2mas/src/migrate.mts
```diff
@@ -119,9 +119,9 @@ export async function migrate(): Promise<void> {
     fatals += 1;
   }
 
-  function makeUuid<T>(): UUID<T> {
-    return id128.Uuid4.fromRaw(
-      id128.UlidMonotonic.generate().toRaw(),
```
sandhose, line 124: I switched from UlidMonotonic because:

 - there is a very low chance of collision on the same timestamp with the random part 
 - UlidMonotonic is buggy when going "back in time"

PR #2234, opened 2024-01-04
Allow `config dump` to generate to a file directly instead of stdout #2146

dklimpel:
It is a little bit copy&paste from
- #2146

Writing the dump to the file system is only consequent.

I am not very familiar with rust to bring it in a seperate function for deduplicated code.

### DIFF
--- a/crates/cli/src/commands/config.rs
+++ b/crates/cli/src/commands/config.rs
@@ -90,7 +90,13 @@ pub(super) struct Options {
 #[derive(Parser, Debug)]
 enum Subcommand {
     /// Dump the current config as YAML
-    Dump,
+    Dump {
+        /// The path to the config file to dump
+        ///
+        /// If not specified, the config will be written to stdout
+        #[clap(short, long)]
+        output: Option<Utf8PathBuf>,
+    },
 
     /// Check a config file
     Check,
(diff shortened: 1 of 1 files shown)


dklimpel:
I am not sure if the failing pipeline is my fault:

```
error[vulnerability]: Some Ref methods are unsound with some type parameters
    ┌─ /github/workspace/Cargo.lock:[54](https://github.com/matrix-org/matrix-authentication-service/actions/runs/7408315024/job/20156268314?pr=2234#step:4:55)3:1
    │
543 │ zerocopy 0.7.26 registry+https://github.com/rust-lang/crates.io-index
    │ --------------------------------------------------------------------- security vulnerability detected
    │
    = ID: RUSTSEC-2023-0074
    = Advisory: https://rustsec.org/advisories/RUSTSEC-2023-0074
    = The `Ref` methods `into_ref`, `into_mut`, `into_slice`, and `into_slice_mut` are unsound
      and may allow safe code to exhibit undefined behavior when used with `Ref<B, T>` where `B`
      is [`cell::Ref`](https://doc.rust-lang.org/core/cell/struct.Ref.html) or
      [`cell::RefMut`](https://doc.rust-lang.org/core/cell/struct.RefMut.html). Note that these
      methods remain sound when used with `B` types other than `cell::Ref` or `cell::RefMut`.
```

sandhose (approved):

PR #2264, opened 2024-01-22
Add a feature for native-tls support.

pixlwave:
On iOS we're using the Rust SDK's native-tls feature to depend on the system for TLS, however when running a `cargo tree` I noticed that we're also pulling in rustls and tracked it back to the OIDC client crate. This PR allows the choice between rustls or native-tls by introducing a second feature.

### DIFF
--- a/crates/oidc-client/Cargo.toml
+++ b/crates/oidc-client/Cargo.toml
@@ -21,6 +21,13 @@ hyper = [
     "dep:tower-http",
     "tower/limit",
 ]
+hyper-tls = [
+    "dep:http-body",
+    "dep:hyper",
+    "dep:hyper-tls",
+    "dep:tower-http",
+    "tower/limit",
+]
 keystore = ["dep:mas-keystore"]
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


crates/oidc-client/Cargo.toml
pixlwave, line 1: I wasn't sure if it was better to make a copy of the `hyper` feature with the TLS dep changed, or to pull out `[hyper-]rustls` into a `hyper-rustls` feature and make the new addition a `hyper-native-tls` feature?

pixlwave: Also, from what I can tell, having rustls enabled by default to avoid a breaking change means that to use native TLS the dependency would need to be configured like so:
```
mas-oidc-client = { version = "0.7.0", default-features = false, features = ["hyper-tls", "keystore"], optional = true }
```
Which doesn't seem great to me if new default features are added but I think this is a limitation of how cargo works?

pixlwave:
Errr CI is complaining about

```
error[banned]: crate 'native-tls = 0.2.11' is explicitly banned
   ┌─ /github/workspace/deny.toml:56:14
   │
56 │     { name = "native-tls" },
   │              ^^^^^^^^^^^^ banned here
   │
```

sandhose:
@pixlwave it turns out, rustls is optional in mas-oidc-client. The Rust SDK is just enabling the default features, which includes the default HTTP client, but actually doesn't use this. I think @zecakeh is working on removing this from the Rust SDK dependency tree

pixlwave:
Oh nice, that's great. Also sorry for opening this and then completely forgetting to come back and look at the failures.

PR #2317, opened 2024-02-06
syn2mas: Skip access tokens that don't have a device ID

jaywink:
These can't be migrated. They are likely enirely valid tokens created by puppeting a user via the Synapse admin api. Treating them as fatal errors will make it impossible to migrate any host that has ever used this admin API feature.

### DIFF
--- a/docs/setup/migration.md
+++ b/docs/setup/migration.md
@@ -8,11 +8,10 @@ Features that are provided to support this include:
 
 - Ability to import existing password hashes from Synapse
 - Ability to import existing sessions and devices
-- Ability to import existing access tokens
+- Ability to import existing access tokens linked to devices (ie not including short-lived admin puppeted access tokens)
 - Ability to import existing upstream IdP subject ID mappings
 - Provides a compatibility layer for legacy Matrix authentication
 
-If
 There will be tools to help with the migration process itself. But these aren't quite ready yet.
 
 ## Preparing for the migration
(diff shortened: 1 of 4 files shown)


sandhose (approved):
This change makes sense to me, although I wonder if we should mention that in the migration docs? At least it warns about it in the advisor, so it's probably fine for now 👍

jaywink:
> This change makes sense to me, although I wonder if we should mention that in the migration docs? At least it warns about it in the advisor, so it's probably fine for now 👍

Have tried to clarify in the docs :+1:

docs/setup/migration.md
```diff
@@ -8,11 +8,10 @@ Features that are provided to support this include:
 
 - Ability to import existing password hashes from Synapse
 - Ability to import existing sessions and devices
-- Ability to import existing access tokens
+- Ability to import existing access tokens linked to devices (ie not including short-lived admin puppeted access tokens)
 - Ability to import existing upstream IdP subject ID mappings
 - Provides a compatibility layer for legacy Matrix authentication
 
-If
```
jaywink, line 15: Random stray `If` removed while touching the file. Possibly there was a sentence being written here at some point?

sandhose: Probably :D

I think I did notice it at some point, told Hugh about it and we both forgot to remove it 🙈

PR #2325, opened 2024-02-07
Make the user agree to T&C during registration & record which document they agreed to

### LINKED ISSUE #22: Instance terms of services and privacy policy
The user should agree to some terms of services before using the auth service. This aims to replace the existing (broken) `m.login.terms`

sandhose:
Fixes #22 

If `branding.tos_uri` is set in the config, this adds a mandatory checkbox during registration.
It then records which the URL of the TOS that the user agreed to in a `user_terms` table.

Note that this doesn't force the user to agree to TOS if they change or anything like that, but the DB schema should be flexible enough to implement that in the future.

<table>
<tr>
<td>

image

</td>
<td>

image

</td>
</tr>
</table>

### DIFF
--- a/crates/cli/src/commands/server.rs
+++ b/crates/cli/src/commands/server.rs
@@ -140,6 +140,7 @@ impl Options {
         );
 
         let site_config = SiteConfig {
+            tos_uri: config.branding.tos_uri.clone(),
             access_token_ttl: config.experimental.access_token_ttl,
             compat_token_ttl: config.experimental.compat_token_ttl,
         };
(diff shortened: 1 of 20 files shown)


americanrefugee:
Link to the updated design in Figma:
[https://www.figma.com/file/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?type=design&node-id=4770-13490&mode=design](https://www.figma.com/file/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?type=design&node-id=4770-13490&mode=design)

sandhose:
@americanrefugee Thanks, I updated the spacing and wording. The weird spacing around the link is a compound-web bug

PR #2351, opened 2024-02-12
Replace jotai-urql with just urql

sandhose:
This was unnecessarily complicated.
This replaces everything which used jotai-urql with just urql calls.

Next step is to replace the custom jotai-based router with tanstack router, but I prefer to do that in a second PR to make it easier to review.

For reviewers, most interesting changes are:

 - in `src/app.tsx`, the app root now loads the current user ID to allow giving a non-nullable user ID down the component tree, and avoid having a "is the user ID null" logic in every route
 - for error handling, I prefer to just throw if `result.error` is set and handle the error in the nearest error boundary
 - annoyingly, `urql` doesn't know it runs in suspense mode in its typings, so `result.data` is nullable, which should never happen unless `result.error` is set. So there is a lot of:
   ```ts
   const [result] = useQuery(...);
   if (result.error) throw result.error;
   if (!result.data) throw new Error(); // Suspense mode is enabled
   ```
 - the pagination logic was probably the biggest change. See `src/pagination.tsx` for the hooks, and `src/components/BrowserSessionList.tsx` for an example usage


t3chguy (approved):
Looks sane

PR #2359, opened 2024-02-15
Replace Jotai with @tanstack/router

### LINKED ISSUE #1713: Improve UI for device-less sessions
Those are created e.g. from the admin console.

sandhose:
This replaces Jotai completely with @tanstack/router.

Fixes #1713

This makes the routing much simpler to understand, as it uses a file-based structure to define routes.
It makes defining error boundaries and not found boundaries way easier.
It makes it way easier to pre-load data before changing pages

The nested routing should help making the individual session view a modal on top of the list.

The main thing that changes are the URLs.

 - `/sessions-overview` -> `/sessions`
 - `/browser-sessions` -> `/sessions/browsers`
 - `/session/:deviceId` -> `/sessions/:id` (note that we now use the session ID here, not the device ID)
 - `/devices/:deviceId` now redirects to the corresponding `/sessions/:id` page

A few points of interest when looking at the code:

 - `src/routes/__root.tsx` has the MSC2965 ?action parameter handling
 - `src/routes/_account.tsx`: for pages which uses the user greeting + tabs, this "ghost" page is used to define this common layout, but without applying it to all pages.
 - `test-utils/router.tsx` has a test helper to create a dummy router with just an index route, which renders whatever children you give it


t3chguy (approved):
Looks sane

PR #2362, opened 2024-02-15
Update the email list to match the new design

sandhose:
Design: https://www.figma.com/file/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?type=design&node-id=4366-39954&mode=design&t=ewAWHeheVaay8GWg-4

New implementation:
image

I also implemented the generic dialog which is mostly copied from the Menu in compound, and should probably move the implementation to compound later


t3chguy (approved):
Looks sane

PR #2403, opened 2024-02-27
Make sure the dummy router finished loading in the test environment

sandhose:
I can't figure out why it was the case, maybe a bug in @tanstack/router, but sometimes the router would render an empty `<div>` instead of its children.
This adds a `router.load()` call in the router test utility, and exports a `routerReady` promise which resolves when the router is ready to be used.

The big diff is because we had empty snapshots because of this. 🤦

### DIFF
--- a/frontend/src/components/CompatSession.test.tsx
+++ b/frontend/src/components/CompatSession.test.tsx
@@ -14,6 +14,7 @@
 
 // @vitest-environment happy-dom
 
+import { TooltipProvider } from "@vector-im/compound-web";
 import { create } from "react-test-renderer";
 import { Provider } from "urql";
 import { describe, expect, it, beforeAll } from "vitest";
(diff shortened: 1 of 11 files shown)


frontend/src/components/OAuth2Session.test.tsx
```diff
@@ -48,6 +48,7 @@ describe("<OAuth2Session />", () => {
 
   const finishedAt = "2023-06-29T03:35:19.451292+00:00";
 
+  beforeAll(async () => await routerReady);
```
t3chguy, line 51: This can probably live in `test-utils/router` - so merely importing it would do it - if that isn't too magical for you

sandhose: You're right that works. Just did that

t3chguy (approved):

PR #2405, opened 2024-02-27
Fix dialogs being blurry and not scrollable

sandhose:
This positions dialogs with a flexbox layout, which avoids the weird bluriness due to `transform: translate` and allows the dialog to be scrollable.

Story to try out: https://quenting-frontend-fix-modal.matrix-authentication-service-docs.pages.dev/storybook/?path=/story/ui-dialog--long-text


t3chguy (approved):
Looks like CSS

PR #2425, opened 2024-02-29
Add Podman support to policies Makefile

AndrewFerr:
These are tweaks that I had to apply in order to build `policies` with Podman on an SELinux-enabled Fedora Workstation system.


sandhose:
Thanks!

PR #2433, opened 2024-03-01
Warn that Element Web may need .well-known config

AndrewFerr:
Worth mentioning for administrators of self-hosted Element Web clients.


sandhose:
I believe this isn't true anymore, as Element Web will now use the /auth_issuer C-S API endpoint to discover the issuer

Issue #2147, opened 2023-11-24
Improve the session detail view


sandhose:
Currently looks like this:

<img src="https://github.com/matrix-org/matrix-authentication-service/assets/1549952/9bc28d47-03e6-4f67-b8cf-c16d67bed398" width="300" />

[Latest designs](https://www.figma.com/file/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?type=design&node-id=4350-20939&mode=design&t=umcQytfGtiOIePzE-4)

Note on the implementation: making it a modal might be tricky as we need to be able to deep link to it. A first step would be to implement this in its own page, without a modal.

The scope list should be consistent with what is done on the server-side template for the consent page: https://github.com/matrix-org/matrix-authentication-service/blob/8f96427c47f048b99f44c7deaf5fdcfdcd98cb38/templates/components/scope.html

hughns:
We need a design for this.

daniellekirkwood:
@hughns , do you have a path to design support or do we need to escalate that ?

pmaier1:
Designs here https://www.figma.com/file/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?type=design&node-id=4344-10620&mode=design&t=fYw0aza1SK7JICbo-4

langleyd:
@t3chguy You mentioned you were blocked on this. I believe you said we didn't have all the data to render the session detail as presented in the designs.

I think it was that device name would be equal to the device id, so that the list would not be actionable for the user. (Similar to this element [web issue](https://github.com/element-hq/element-web/issues/26876)?)

Would you mind confirming so this is actionable for @sandhose ?

t3chguy:
Yes, I believe this was discussed either in the call or in the room, that MAS doesn't have a concept of editable device names at this time, and at the API level it seems that the name for certain session types is entirely unavailable which means that on the frontend the [matrix] Device ID is used instead.

sandhose:
@t3chguy This should be unblocked now that we have more user-agent infos on the sessions. It's not a user-editable session name, but it should be enough to implement something sensible

daniellekirkwood:
there are a couple of PRs here, did we finish this task?

t3chguy:
@daniellekirkwood I think so

Design:
![image](https://github.com/user-attachments/assets/886f401a-6aff-4417-8ff8-3092a4a1232c)

Reality:
[...]

PR #2439, opened 2024-03-04
Update session details styles to closer match latest Figma

t3chguy:
For https://github.com/matrix-org/matrix-authentication-service/issues/2147

image
image
image


frontend/src/components/SessionDetail/CompatSessionDetail.tsx
```diff
@@ -124,6 +92,10 @@ const CompatSessionDetail: React.FC<Props> = ({ session }) => {
       <SessionHeader to="/sessions">{data.deviceId || data.id}</SessionHeader>
       <SessionDetails
         title={t("frontend.compat_session_detail.session_details_title")}
+        deviceId={data.deviceId}
+        signedIn={parseISO(data.createdAt)}
+        lastActive={data.lastActiveAt ? parseISO(data.lastActiveAt) : undefined}
+        ipAddress={data.lastActiveIp ?? undefined}
```
sandhose, line 98: On the consent screen for compat sessions, we "fake" a scope list to have something ~coherent ; we could do the same here?

https://github.com/matrix-org/matrix-authentication-service/blob/e5a88622db81c90315d2942664c664ee795da604/templates/pages/sso.html#L34

frontend/src/components/SessionDetail/SessionDetails.module.css
```diff
@@ -13,28 +13,51 @@
...
+  display: flex;
+  flex-direction: column;
+  gap: var(--cpd-space-scale);
+  border-radius: var(--cpd-space-5x);
+  overflow: hidden;
+}
+
+.scope {
+  background: var(--cpd-color-bg-subtle-secondary);
+  padding: 12px 20px 12px 20px;
```
sandhose, line 53:
```suggestion
  padding: var(--cpd-space-3x) var(--cpd-space-5x);
```

frontend/src/components/SessionDetail/SessionDetails.module.css
```diff
@@ -13,28 +13,51 @@
...
+  display: flex;
+  align-items: center;
+  gap: 12px;
```
sandhose, line 56:
```suggestion
  gap: var(--cpd-space-3x);
```

frontend/src/components/SessionDetail/SessionDetails.module.css
```diff
@@ -13,28 +13,51 @@
...
-    display: flex;
-    flex-direction: column;
-    margin-top: var(--cpd-space-1x);
-    gap: var(--cpd-space-1x);
+.wrapper {
+  display: flex;
+  flex-wrap: wrap;
+  gap: 16px;
+  margin-bottom: 16px;
+  margin-top: 32px;
```
sandhose, line 21:
```suggestion
  gap: var(--cpd-space-4x);
  margin-bottom: var(--cpd-space-4x);
  margin-top: var(--cpd-space-8x);
```

sandhose (changes requested):
The generic session details component still feels messy, but I'm fine with refactoring that later.

Left a few comments but other than that LGTM

frontend/src/components/SessionDetail/CompatSessionDetail.tsx
t3chguy: Yes, but I'd argue that should be a separate PR as the design doesn't explicitly state to do this only showing one comp.

frontend/src/components/SessionDetail/SessionDetails.module.css
t3chguy: Ah yes, I forgot to go back to the spacing vars, my bad

t3chguy:
> The generic session details component still feels messy, but I'm fine with refactoring that later.

That component gets deleted by the end of the PR, no?

sandhose:
> That component gets deleted by the end of the PR, no? Or do you mean the details prop behaviour?

I meant that this component has a lot of logic in it. Sure, it avoids repeating stuff across the three sessions, but I prefer the approach of composing different components like on the session card: https://github.com/matrix-org/matrix-authentication-service/blob/6eb6209bd88e526d75441acdef07626dabe3a4fa/frontend/src/components/BrowserSession.tsx#L119-L163

t3chguy:
> it avoids repeating stuff across the three sessions

I think this is key to a consistent coherent design, otherwise you have to manually sync the 3 session components for copy and order to keep them consistent with each other. Personally not a fan, but at the end of the day this is your project

sandhose (approved):

sandhose:
I'm fine with leaving it like that for now 👍

PR #2470, opened 2024-03-12
Add account management URL for clients

zecakeh:
Introduces types and methods to access the account management URL for clients.

### DIFF
--- a/crates/oidc-client/src/requests/account_management.rs
+++ b/crates/oidc-client/src/requests/account_management.rs
@@ -1,0 +1,127 @@
+// Copyright 2024 Kévin Commaille.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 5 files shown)


crates/oidc-client/src/requests/account_management.rs
```diff
@@ -0,0 +1,127 @@
...
+use serde::Serialize;
+use serde_with::skip_serializing_none;
+use url::Url;
+
+/// An account management action that a user can take, including a device ID for
+/// the actions that support it.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+#[serde(tag = "action")]
+#[non_exhaustive]
+pub enum AccountManagementActionFull {
```
sandhose, line 28: There is some duplication with this struct but that's probably fine for now
https://github.com/matrix-org/matrix-authentication-service/blob/2e8d4da80025c08779b84d971370442dde735b53/crates/router/src/endpoints.rs#L455-L481

sandhose (approved):
Thanks!

Issue #2265, opened 2024-01-22
Improve the session view


t3chguy:
https://www.figma.com/file/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?type=design&node-id=4344-10620&mode=design&t=fYw0aza1SK7JICbo-4

![image](https://github.com/matrix-org/matrix-authentication-service/assets/2403652/a852d95e-a10d-4618-82b0-d10f1aba8dab)

sandhose:
Some of this has been done in #2393

matrixbot:
For your information, this issue has been copied over to the Element fork of `matrix-authentication-service`: <https://github.com/element-hq/matrix-authentication-service/issues/2265>

PR #2478, opened 2024-03-13
Render browser icon for compat & browser sessions

t3chguy:
image

For https://github.com/matrix-org/matrix-authentication-service/issues/2265


sandhose (approved):
Sorry that you had to deal with the assets loading stuff but thanks for this!

PR #2574, opened 2024-04-02
Create the AccountManagementError enum

zecakeh:
Otherwise it forces dependant crates to add serde_urlencoded as a dependency just to be able to forward the error type.

We don't use that crate directly in matrix-sdk, so it is a bit annoying to have to add that dependency just to forward the error.


sandhose (approved):

sandhose:
That makes sense, thanks!

PR #2653, opened 2024-04-25
add helm chart for MAS deployments

phk-nord:
This PR intends to add a helm chart.
Although this does not automate the setup and configuration of the authentication service it may be something people might want to have as basis


helm/configs/config.yaml
```diff
@@ -0,0 +1,298 @@
...
+      # List of resources to serve
+      resources: {{- range .Values.config.listeners.resources }}
+      - name: {{ .name }}
+      {{- if .path }}
+        path: {{ .path -}}
+      {{- end }}
+      {{- if .playground }}
+        playground: {{ .playground }}
+      {{- end }}
+      {{- end }}
```
sandhose, line 23: If anyway the chart exposes everything on a single port (which is fine!), I wouldn't expose the various resources in the values, but rather flags like `graphqlPlayground`, `metrics`, etc. and change that section accordingly

helm/configs/config.yaml
```diff
@@ -0,0 +1,298 @@
...
+    protocol: {{ .Values.telemetry.useJaeger.protocol }}
+    endpoint: {{ .Values.telemetry.useJaeger.endpoint }}
+    username: {{ .Values.telemetry.useJaeger.username }}
+    password: {{ .Values.telemetry.useJaeger.password }}
+    agent_host: {{ .Values.telemetry.useJaeger.agent_host }}
+    agent_port: {{ .Values.telemetry.useJaeger.agent_port }}
+    {{- end }}
+    {{- if .Values.telemetry.useZipkin.enabled }}
+    collector_endpoint: {{ .Values.telemetry.useZipkin.collector_endpoint }}
+    {{- end }}
```
sandhose, line 164: The jaeger and zipkin exporters were removed in 0.9.0

helm/configs/config.yaml
```diff
@@ -0,0 +1,298 @@
...
+  command: {{ .Values.email.command }}
+  {{- end }}
+  mode: {{ .Values.email.mode }}
+  port: {{ .Values.email.port }}
+  username: {{ .Values.email.username }}
+  password: {{ .Values.email.password }}
+{{- end }}
+
+  # Settings related to upstream OAuth 2.0/OIDC providers. This section must be synced to the database using the config sync command.
+  # https://matrix-org.github.io/matrix-authentication-service/usage/cli/config.html#config-sync---prune---dry-run
```
sandhose, line 192: This is not the case anymore, as they get synced automatically on startup now

helm/templates/configmap.yaml
```diff
@@ -0,0 +1,33 @@
...
+        "authorization_endpoint": {{ .Values.wellKnown.authEnd | quote}},
+        "token_endpoint": {{ .Values.wellKnown.tokenEnd | quote}},
+        "jwks_uri": {{ .Values.wellKnown.jwks | quote}},
+        "registration_endpoint": {{ .Values.wellKnown.registerEnd | quote}},
+        "introspection_endpoint": {{ .Values.wellKnown.introEnd | quote}},
+        "response_types_supported": ["code"],
+        "code_challenge_methods_supported": ["S256"],
+        "grant_types_supported": ["authorization_code"],
+        "revocation_endpoint": {{ .Values.wellKnown.revokeEnd | quote}}
+        }`
```
sandhose, line 32: This really should be served by MAS directly, as this document might change with the versions and configuration

helm/values.yaml
```diff
@@ -0,0 +1,409 @@
...
+  #   memory: 128Mi
+
+livenessProbe:
+  httpGet:
+    path: /oauth2/keys.json
+    port: http
+readinessProbe:
+  httpGet:
+    path: /oauth2/keys.json
+    port: http
```
sandhose, line 89: There is a dedicated healthcheck endpoint on /health, that you could mount on a seperate port

helm/values.yaml
```diff
@@ -0,0 +1,409 @@
...
+affinity: {}
+
+wellKnown:
+  issuer: https://auth.example.com/
+  authEnd: https://auth.example.com/authorize
+  tokenEnd: https://auth.example.com/oauth2/token
+  jwks: https://auth.example.com/oauth2/keys.json
+  registerEnd: https://auth.example.com/oauth2/registration
+  introEnd: https://auth.example.com/oauth2/introspect
+  revokeEnd: https://auth.example.com/oauth2/revoke
```
sandhose, line 133: Why would an end user need to configure this?

helm/values.yaml
```diff
@@ -0,0 +1,409 @@
...
+        path: /usr/local/share/mas-cli/assets/
+    binds:
+      # use either address, host/port (hp) or socket
+      type:
+        address: 
+          enabled: false
+        hp:
+          enabled: true
+        socket: 
+          enabled: false
```
sandhose, line 159: why would you need to change this? also `hp` feels unnecessarily obscure

helm/values.yaml
```diff
@@ -0,0 +1,409 @@
...
+  redirect_uris:
+    - http://localhost:1234/callback
+  # Public client
+  pub_id: 0000000000000000000SYNAPSX
+  pubAuthMethod: none
+
+secrets:
+  # Encryption secret (used for encrypting cookies and database fields)
+  # This must be a 32-byte long hex-encoded key
+  encrypt:
```
sandhose, line 217: Can't we have helm automatically generate something here and put that in a secret?

helm/values.yaml
```diff
@@ -0,0 +1,409 @@
...
+      key: |-
+
+    - kid: keyid#2
+      key: |-
+
+    - kid: keyid#3
+      key: |
+
+    - kid: keyid#4
+      key: |
```
sandhose, line 237: Same here: you could generate a RSA key and a ECDSA P-256 key with Helm and that would be enough

helm/values.yaml
sandhose, line 1: The configuration keys feel quite inconsistent. Sometimes you use `camelCase`, sometimes `snake_case` ; sometimes it's a 1-1 mapping of what is in the MAS config, sometimes it's another term used…

sandhose (commented):
Thanks for this! I've really quickly skimmed through, this isn't a full review.

I'm not yet sure whether this should live in this repository or somewhere else, as I'm not sure I have enough time to maintain this and make sure it doesn't break. Maybe we could do something like synapse does and have it in a `contrib/` directory, to make it explicit as it being an external contribution, but not necessarily well tested?

helm/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v2
+name: mas
+description: A Helm chart for matrix-authentication-service
+type: application
+version: 0.1.0
+appVersion: "1.16.0"
```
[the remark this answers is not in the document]
phk-nord: yes, thats a remain from `helm create`

helm/configs/config.yaml
phk-nord: which listener make sense to expose by default and which should always be optional besides `graphqlPlayground` and `metrics` ?

helm/configs/config.yaml
```diff
@@ -0,0 +1,298 @@
...
+      {{- end }}
+      {{- if .Values.config.listeners.binds.type.hp.enabled }}
+        # Second option: listen on the given host and port combination
+        - host: {{ .Values.config.listeners.binds.host.name }}
+          port: {{ .Values.config.listeners.binds.host.port }}
+      {{- end }}
+      {{- if .Values.config.listeners.binds.type.socket.enabled }}
+        # Third option: listen on the given UNIX socket
+        - socket: {{ .Values.config.listeners.binds.socket }}
+      {{- end }}
```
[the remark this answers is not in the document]
phk-nord: > Is it useful to bind on other host/ports?
almost never, no. There are obscure one node use cases, however you're right, this should be something that covers a regular use case instead

helm/values.yaml
phk-nord: I did try to use the original MAS config, as I am a fan of referring to the regular documentation rather than just doing something entirely else. Sometimes however my mind gets off track. the `hp` for host+port was indeed quite odd

helm/configs/config.yaml
```diff
@@ -0,0 +1,281 @@
...
+      {{- if .path }}
+        path: {{ .path -}}
+      {{- end }}
+      {{- if .playground }}
+        playground: {{ .playground }}
+      {{- end }}
+      {{- end }}
+      - name: health
+        path: /health
+      {{- if .Values.telemetry.metrics.enabled }}
```
gabreal, line 26: what is the idea behind putting some config values below `.Values.config` and others like this not?

helm/configs/config.yaml
```diff
@@ -0,0 +1,281 @@
...
+  path: {{ .Values.templates.path }}
+  assets_manifest: {{ .Values.templates.assets }}
+{{- end }}
+
+  # List of OAuth 2.0/OIDC clients and their keys/secrets. Each client_id must be a ULID.
+clients:
+  # Confidential client
+  - client_id: {{ .Values.clients.client_id }}
+    client_auth_method: {{ .Values.clients.client_auth_method }}
+    client_secret: {{ .Values.clients.client_secret }}
```
gabreal, line 84: i understand that secrets handling mostly depends on custom integrations, but having secrets in a configmap is not so typical. [anance's synapse chart](https://gitlab.com/ananace/charts/-/blob/master/charts/matrix-synapse/templates/deployment.yaml?ref_type=heads#L69) for instance puts them into separate config files from secrets. it has pros and cons.

helm/README.md
```diff
@@ -0,0 +1,41 @@
+# Oddities with MAS
+
+## general
+
+After initialising the pod you need to sync MAS with it's config
+
+```bash
+kubectl exec -it -n mas deployments/mas -- mas-cli config sync
```
gabreal, line 8: the sync could be run from an init container. since the migrations and basic sync are now performed automatically on start up we added:
```
    - name: mas-sync-clients
      image: ghcr.io/matrix-org/matrix-authentication-service:0.9.0
      command: ["/usr/local/bin/mas-cli", "config", "sync", "--config=/config/config.yaml", "--prune"]
      volumeMounts:
        - mountPath: /config
          name: config
```

helm/values.yaml
```diff
@@ -0,0 +1,342 @@
...
+# This is a YAML-formatted file.
+# Declare variables to be passed into your templates.
+
+replicaCount: 1
+
+image:
+  repository: ghcr.io/matrix-org/matrix-authentication-service
+  pullPolicy: IfNotPresent
+  # Overrides the image tag whose default is the chart appVersion.
+  tag: "0.9.0"
```
gabreal, line 11: why not use `appVersion` from the chart as default ([ref](https://gitlab.com/ananace/charts/-/blob/master/charts/matrix-synapse/Chart.yaml?ref_type=heads#L6))?

PR #2718, opened 2024-05-03
Relax alphabet restrictions on device IDs

sandhose:
I tested with the following device ID: `AZaz09._!$()*+,;=:@/-'` and checked that
 - the creation in Synapse worked
 - the redirection for `/account?action=org.matrix.session_end&device_id=AZaz09._!$()*+,;=:@/-'` worked
 - the redirection for `/account/devices/AZaz09._!$()*+,;=:@/-'` worked
 - the deletion in Synapse worked

It looks like one exception in the characters added to the alphabet is the `&`, which work for the Synapse creation and deletion, as well as `/account/devices/$id`, but doesn't for `/account/?action=…`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3411,6 +3411,7 @@ dependencies = [
  "tower",
  "tracing",
  "url",
+ "urlencoding",
 ]
 
 [[package]]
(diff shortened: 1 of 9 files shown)


hughns (approved):
Looks good to me

PR #2759, opened 2024-05-13
CAPTCHA verification during registration

### LINKED ISSUE #138: CAPTCHA support
During registration, and optionally other operations. To replace the HS capability such as https://github.com/matrix-org/synapse/blob/develop/docs/CAPTCHA_SETUP.md

Open questions:

 - Which CAPTCHA provider should we support? With #2759, we will support all of:
   - [reCAPTCHA](https://www.google.com/recaptcha/about/) is what we already support in Synapse 
   - [hCaptcha](https://www.hcaptcha.com) for a non-Google alternative, highly requested in Synapse by the community
   - [Cloudflare Turnstile](https://www.cloudflare.com/en-gb/products/turnstile/) might be the most user-friendly?
 - What operations should be protected? Obvious candidates are:
   - Password-based login
   - Password-based registration
   - Password change
   - When adding an email address
 - Where does it fit in the login/registration forms? -> Design suggestion in #2759
 - What configuration knob do we give to administrators?

Relevant spec issue: https://github.com/matrix-org/matrix-spec/issues/295

sandhose:
This adds support for Cloudflare Turnstile, reCAPTCHA and hCAPTCHA validation during self-service user registration.
They all have very similar APIs, so supporting all three isn't very costly.

Fixes #138 

Follow up to do:

 - [x] Make sure the placement/style is right
 - [x] Display user-friendly errors on CAPTCHA failures (needs design)
 - [x] Have a `<noscript>` message to tell users to enable JS
 - [ ] Protect other actions?
 - [x] Document the feature

| reCAPTCHA | Cloudflare Turnstile | hCAPTCHA |
| --- | --- | --- |
| image | image | image |

| `<noscript>` fallback | Failure error message | 
| --- | --- |
| image | image |

### DIFF
--- a/crates/cli/src/commands/server.rs
+++ b/crates/cli/src/commands/server.rs
@@ -145,7 +145,8 @@ impl Options {
             &config.matrix,
             &config.experimental,
             &config.passwords,
-        );
+            &config.captcha,
+        )?;
 
         // Load and compile the templates
         let templates =
--- a/crates/cli/src/commands/templates.rs
+++ b/crates/cli/src/commands/templates.rs
@@ -15,8 +15,8 @@
 use clap::Parser;
 use figment::Figment;
 use mas_config::{
-    BrandingConfig, ConfigurationSection, ExperimentalConfig, MatrixConfig, PasswordsConfig,
-    TemplatesConfig,
+    BrandingConfig, CaptchaConfig, ConfigurationSection, ExperimentalConfig, MatrixConfig,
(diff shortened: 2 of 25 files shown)


americanrefugee:
- Placement of the CAPTCHA widget is fine
- Placement of banner is fine (e.g. JavaScript needs to be enabled)
- However, the error message (e.g. "Verification failed") should go just below the widget / above the Continue button

sandhose:
I fixed the placement of the error message

PR #2779, opened 2024-05-16
Minor drive-by tweaks

reivilibre:
Nothing major at all, just a few suggestions as I was reading some code


sandhose (approved):
All of this makes sense, thanks!

PR #2783, opened 2024-05-17
Merge the `mas_graphql` crate into the `mas_handlers` crate

reivilibre:
This one is to be squash merged, it's only presented as separate non-atomic commits for easier review
(I personally like to have the rename-only operation, as git recognises it, as one commit)


crates/handlers/src/graphql/mod.rs
```diff
@@ -106,9 +123,9 @@ pub fn schema(
         homeserver_connection: Arc::new(homeserver_connection),
         site_config,
     };
-    let state: mas_graphql::BoxState = Box::new(state);
+    let state: crate::graphql::state::BoxState = Box::new(state);
```
sandhose, line 126:
```suggestion
    let state: self::state::BoxState = Box::new(state);
```

or actually import `BoxState` not just in `#[cfg(test)]`?

crates/handlers/src/graphql/mod.rs
```diff
@@ -44,7 +46,22 @@ use rand::{thread_rng, SeedableRng};
...
 use tracing::{info_span, Instrument};
+use ulid::Ulid;
+
+mod model;
+mod mutations;
+mod query;
+mod state;
 
+#[cfg(test)]
+pub use state::BoxState;
```
sandhose, line 57: I don't understand why is this under `cfg(test)`?

crates/handlers/src/graphql/model/oauth.rs
```diff
@@ -21,7 +21,7 @@ use ulid::Ulid;
 use url::Url;
 
 use super::{BrowserSession, NodeType, SessionState, User, UserAgent};
-use crate::{state::ContextExt, UserId};
+use crate::graphql::{state::ContextExt, UserId};
```
sandhose, line 24: I don't have a strong opinion, but maybe we should grab all this from `super::super`? 
```suggestion
use super::super::{state::ContextExt, UserId};
```

crates/handlers/src/graphql/mod.rs
```diff
@@ -44,7 +46,22 @@ use rand::{thread_rng, SeedableRng};
...
+pub use state::State;
```
sandhose, line 58: I really like using `self::` when doing relative imports, so that it doesn't clash with a potential crate called `state`. wdyt?

sandhose (changes requested):
Looks sane overall, thanks!

crates/handlers/src/graphql/mod.rs
reivilibre: It got a warning for being unused when not in tests :)

reivilibre: using it below directly serves as a workaround, so don't need this anymore

crates/handlers/src/graphql/mod.rs
```diff
@@ -338,3 +355,145 @@ pub async fn playground() -> impl IntoResponse {
...
+#[must_use]
+pub fn schema_builder() -> SchemaBuilder {
+    async_graphql::Schema::build(Query::new(), Mutation::new(), EmptySubscription)
+        .register_output_type::<Node>()
+        .register_output_type::<CreationEvent>()
+}
+
+/// The identity of the requester.
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub enum Requester {
```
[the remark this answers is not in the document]
reivilibre: mrmm, I did think as much but mostly got frozen on what to call the extra file.

crates/handlers/src/graphql/model/oauth.rs
sandhose: you're right, that's fine with me

crates/handlers/src/graphql/mod.rs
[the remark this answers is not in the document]
sandhose: I did suspect that :D
Let's postpone that to later™

sandhose (approved):

PR #2794, opened 2024-05-23
Improve errors when MAS contacts the Synapse homeserver

reivilibre:
An example of a failed registration now looks like:

Screenshot_20240523_172645

Before the only detail this gave you was the headline message repeated a few times.

The logs now say

```
2024-05-23T16:26:32.863728Z ERROR http.server.request{otel.kind="server" otel.name="POST /register" network.protocol.name="http" network.protocol.version="1.1" http.request.method="POST" url.path="/register" url.scheme="http" http.route="/register" user_agent.original="Mozilla/5.0 (X11; Linux x86_64; rv:125.0) Gecko/20100101 Firefox/125.0"}:handlers.views.register.post: mas_handlers::views::register: error=Internal error: Failed to query localpart availability from Synapse (Failed to query localpart availability from Synapse

Caused by:
    request failed with status 401 Unauthorized: M_UNKNOWN_TOKEN: Invalid access token passed.)
```

Should be commit-by-commit review friendly

### DIFF
--- a/crates/http/src/ext.rs
+++ b/crates/http/src/ext.rs
@@ -14,7 +14,7 @@
 
 use std::{ops::RangeBounds, sync::OnceLock};
 
-use http::{header::HeaderName, Request, StatusCode};
+use http::{header::HeaderName, Request, Response, StatusCode};
 use tower::Service;
 use tower_http::cors::CorsLayer;
 
@@ -89,17 +89,25 @@ pub trait ServiceExt<Body>: Sized {
         FormUrlencodedRequest::new(self)
     }
 
-    fn catch_http_code<M>(self, status_code: StatusCode, mapper: M) -> CatchHttpCodes<Self, M>
+    /// Catches responses with the given status code and then maps those
+    /// responses to an error type using the provided `mapper` function.
+    fn catch_http_code<M, ResBody, E>(
+        self,
+        status_code: StatusCode,
+        mapper: M,
+    ) -> CatchHttpCodes<Self, M>
     where
-        M: Clone,
+        M: Fn(Response<ResBody>) -> E + Send + Clone + 'static,
     {
         self.catch_http_codes(status_code..=status_code, mapper)
     }
 
-    fn catch_http_codes<B, M>(self, bounds: B, mapper: M) -> CatchHttpCodes<Self, M>
+    /// Catches responses with the given status codes and then maps those
+    /// responses to an error type using the provided `mapper` function.
+    fn catch_http_codes<B, M, ResBody, E>(self, bounds: B, mapper: M) -> CatchHttpCodes<Self, M>
     where
         B: RangeBounds<StatusCode>,
-        M: Clone,
+        M: Fn(Response<ResBody>) -> E + Send + Clone + 'static,
     {
         CatchHttpCodes::new(self, bounds, mapper)
     }
--- a/crates/http/src/ext.rs
+++ b/crates/http/src/ext.rs
@@ -70,6 +70,9 @@ pub trait ServiceExt<Body>: Sized {
         BytesToBodyRequest::new(self)
     }
 
+    /// Adds a layer which collects all the response body into a contiguous
+    /// byte buffer.
+    /// This makes the response type `Response<Bytes>`.
     fn response_body_to_bytes(self) -> BodyToBytesResponse<Self> {
         BodyToBytesResponse::new(self)
     }
--- a/crates/http/src/ext.rs
+++ b/crates/http/src/ext.rs
@@ -111,6 +111,18 @@ pub trait ServiceExt<Body>: Sized {
     {
         CatchHttpCodes::new(self, bounds, mapper)
     }
+
+    /// Shorthand for [`Self::catch_http_codes`] which catches all client errors
+    /// (4xx) and server errors (5xx).
+    fn catch_http_errors<M, ResBody, E>(self, mapper: M) -> CatchHttpCodes<Self, M>
+    where
+        M: Fn(Response<ResBody>) -> E + Send + Clone + 'static,
+    {
+        self.catch_http_codes(
+            StatusCode::from_u16(400).unwrap()..StatusCode::from_u16(600).unwrap(),
+            mapper,
+        )
+    }
 }
 
 impl<S, B> ServiceExt<B> for S where S: Service<Request<B>> {}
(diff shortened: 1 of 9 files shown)


crates/http/src/layers/catch_http_codes.rs
```diff
@@ -45,10 +45,16 @@ impl<S, E> Error<S, E> {
...
+/// A layer that catches responses with the HTTP status codes lying within
+/// `bounds` and then maps the requests into a custom error type using `mapper`.
 #[derive(Clone)]
 pub struct CatchHttpCodes<S, M> {
+    /// The inner service
     inner: S,
+    /// Which HTTP status codes to catch
     bounds: (Bound<StatusCode>, Bound<StatusCode>),
+    /// The function used to convert errors, which must be `Clone` and
+    /// `Fn(Response<ResBody>) -> E + Send + Clone + 'static`.
```
sandhose, line 53: I think `M: Fn(Response<ResBody>) -> E + Send + Clone + 'static` should be understood as

```
M: Fn(Response<ResBody>) -> E
M: Send
M: Clone
M: 'static
```

so you're mentioning Clone twice here?

crates/http/src/ext.rs
```diff
@@ -89,17 +89,25 @@ pub trait ServiceExt<Body>: Sized {
...
+    /// Catches responses with the given status code and then maps those
+    /// responses to an error type using the provided `mapper` function.
+    fn catch_http_code<M, ResBody, E>(
+        self,
+        status_code: StatusCode,
+        mapper: M,
+    ) -> CatchHttpCodes<Self, M>
     where
-        M: Clone,
+        M: Clone + Fn(Response<ResBody>) -> E + Send + Clone + 'static,
```
sandhose, line 100: `Clone` is redundant here, `Fn(Response<ResBody>) -> E` is one trait bound, the list after are other traits associated to `M`, not to the function result

If we wanted to bind traits on the result, we would have bound them to `E`

crates/http/src/ext.rs
```diff
@@ -89,17 +89,25 @@ pub trait ServiceExt<Body>: Sized {
...
     }
 
-    fn catch_http_codes<B, M>(self, bounds: B, mapper: M) -> CatchHttpCodes<Self, M>
+    /// Catches responses with the given status codes and then maps those
+    /// responses to an error type using the provided `mapper` function.
+    fn catch_http_codes<B, M, ResBody, E>(self, bounds: B, mapper: M) -> CatchHttpCodes<Self, M>
     where
         B: RangeBounds<StatusCode>,
-        M: Clone,
+        M: Clone + Fn(Response<ResBody>) -> E + Send + Clone + 'static,
```
sandhose, line 110: Same

crates/http/src/ext.rs
```diff
@@ -111,6 +111,18 @@ pub trait ServiceExt<Body>: Sized {
     {
         CatchHttpCodes::new(self, bounds, mapper)
     }
+
+    /// Shorthand for [`Self::catch_http_codes`] which catches all client errors
+    /// (4xx) and server errors (5xx).
+    fn catch_http_errors<M, ResBody, E>(self, mapper: M) -> CatchHttpCodes<Self, M>
+    where
+        M: Clone + Fn(Response<ResBody>) -> E + Send + Clone + 'static,
```
sandhose, line 119: Same

templates/pages/error.html
```diff
@@ -43,9 +43,8 @@ <h1 class="title">{{ _("error.unexpected") }}</h1>
 
     {% if details %}
       <hr />
-      <pre>
-        <code class="font-mono whitespace-pre-wrap break-all">{{ details }}</code>
-      </pre>
+      {# caution: do not introduce whitespace between <pre> and <code> #}
```
sandhose, line 46: Good catch, I've noticed that but it never bothered me enough to actually fix it :D

crates/matrix-synapse/src/error.rs
```diff
@@ -0,0 +1,72 @@
...
+}
+
+impl Error for HomeserverError {
+    fn description(&self) -> &str {
+        if let Some(matrix_error) = &self.matrix_error {
+            &matrix_error.error
+        } else {
+            "Homeserver Error"
+        }
+    }
```
sandhose, line 43: `description` has been deprecated for a while, it's delegated to Debug/Display now

crates/matrix-synapse/src/lib.rs
```diff
@@ -14,16 +14,28 @@
 
 #![allow(clippy::blocks_in_conditions)]
 
+use anyhow::{bail, Context};
+use error::catch_homeserver_error;
```
sandhose, line 18: Can you use `self::` when doing relative imports? I don't like that they get mixed up with libraries

crates/matrix-synapse/src/lib.rs
```diff
@@ -189,25 +207,57 @@ impl HomeserverConnection for SynapseConnection {
         err(Display),
     )]
     async fn is_localpart_available(&self, localpart: &str) -> Result<bool, Self::Error> {
+        #[derive(Deserialize)]
+        struct UsernameAvailableResponse {
+            available: bool,
+        }
```
sandhose, line 213: The other response structures are defined at the top of the file, for consitency could you move that up there?

sandhose (changes requested):

crates/matrix-synapse/src/error.rs
reivilibre: I think you still have to implement it? Maybe I should have just hardcoded a "don't description() me, you silly!" on it but ehh

crates/matrix-synapse/src/lib.rs
reivilibre: seems like this might be fixable by using `rust-analyzer.imports.prefix = "self"` in the rust-analyzer config

crates/matrix-synapse/src/lib.rs
reivilibre: sure. It's just a personal style I like for these single-use structs, to make them function-local, but yeah indeed there are others here so let's do that

crates/matrix-synapse/src/error.rs
sandhose: looks like the stdlib already does that for you :)

https://doc.rust-lang.org/stable/src/core/error.rs.html#110

sandhose (approved):
woop, forgot to get back to this, sorry

If you could just remove the `Error::description` implementation, otherwise lgtm

crates/matrix-synapse/src/error.rs
reivilibre: huh ok, weird that it added it (rust analyser?) for implementation then. Maybe it just always does that and I'm used to IntelliJ-Rust too much

PR #2808, opened 2024-05-29
Allow parameters in UserInfo endpoint's response's content-type

zecakeh:
`application/json; charset=utf8` should be as valid as `application/json`, so we only check the "essence" of the content type.

### DIFF
--- a/crates/oidc-client/src/requests/userinfo.rs
+++ b/crates/oidc-client/src/requests/userinfo.rs
@@ -19,8 +19,8 @@
 use std::collections::HashMap;
 
 use bytes::Bytes;
-use headers::{Authorization, HeaderMapExt, HeaderValue};
-use http::header::{ACCEPT, CONTENT_TYPE};
+use headers::{Authorization, ContentType, HeaderMapExt, HeaderValue};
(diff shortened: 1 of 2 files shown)


crates/oidc-client/src/requests/userinfo.rs
```diff
@@ -98,16 +99,17 @@ pub async fn fetch_userinfo(
         .call(userinfo_request)
         .await?;
 
-    let content_type = userinfo_response
+    let content_type: Mime = userinfo_response
         .headers()
         .get(CONTENT_TYPE)
```
zecakeh, line 104: Maybe we should use [`.typed_try_get()`](https://docs.rs/headers/latest/headers/trait.HeaderMapExt.html#tymethod.typed_try_get) here instead?

sandhose: `mas-oidc-client` already depends on `headers`, so yeah let's use this instead 👍

zecakeh: Done.

sandhose (approved):

PR #2820, opened 2024-06-04
Add a `setPassword` GraphQL mutation for setting a user's password

### LINKED ISSUE #2197: API for setting password
GraphQL mutation to set/change password.

For a regular user require existing password to be provided (as well as new password).

For an admin no existing password is needed.

reivilibre:
Closes #2197.

### DIFF
--- a/crates/handlers/src/graphql/mutations/user.rs
+++ b/crates/handlers/src/graphql/mutations/user.rs
@@ -19,6 +19,7 @@ use mas_storage::{
     user::UserRepository,
 };
 use tracing::{info, warn};
+use zeroize::Zeroizing;
 
 use crate::graphql::{
     model::{NodeType, User},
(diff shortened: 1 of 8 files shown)


crates/handlers/src/graphql/mutations/user.rs
```diff
@@ -385,4 +446,104 @@ impl UserMutations {
 
         Ok(AllowUserCrossSigningResetPayload::Allowed(user))
     }
+
+    async fn set_password(
```
sandhose, line 450: Could you add a docstring here so that it surfaces in the schema?

crates/handlers/src/graphql/mutations/user.rs
```diff
@@ -385,4 +446,104 @@ impl UserMutations {
...
+            if !state.site_config().password_change_allowed || !password_manager.is_enabled() {
+                return Ok(SetPasswordPayload {
+                    status: SetPasswordStatus::PasswordChangesDisabled,
+                });
+            }
+
+            let Some(active_password) = repo.user_password().active(&user).await? else {
+                // The user has no current password, so can't verify against one.
+                // TODO It may be desirable to let the user set a password without any other
+                // verification instead.
```
sandhose, line 499: This is fine for now, we'll reevaluate once we start thinking about local password + social login use cases (aka matrix.org)

crates/handlers/src/graphql/mutations/user.rs
```diff
@@ -385,4 +446,104 @@ impl UserMutations {
...
+                    status: SetPasswordStatus::NoCurrentPassword,
+                });
+            };
+
+            let Some(current_password_attempt) = input.current_password else {
+                // TODO Maybe this is the wrong error code for when you try to change
+                // your own password without supplying the current password?
+                return Ok(SetPasswordPayload {
+                    status: SetPasswordStatus::NotAllowed,
+                });
```
sandhose, line 511: Considering we're throwing a big "unauthorized" error in case you're trying to edit another user than yours, it might be okay to throw in a case where you're not calling the API the way you're supposed to?

crates/handlers/src/graphql/mutations/user.rs
```diff
@@ -199,6 +200,66 @@ impl AllowUserCrossSigningResetPayload {
...
+    current_password: Option<String>,
+
+    /// The new password for the user.
+    new_password: String,
+}
+
+/// The return type for the `setPassword` mutation.
+#[derive(Description)]
+struct SetPasswordPayload {
+    status: SetPasswordStatus,
```
sandhose, line 222: In other mutations, we return the object which changed, so that the app can directly things which changed in the mutation without doing another round trip.
Example of such a thing: https://github.com/matrix-org/matrix-authentication-service/blob/a63ba4cb8f2e9d756ecec0323e46877ee474f239/crates/graphql/src/mutations/user.rs#L153-L172

buuut since we're not yet exposing whether the user has a password or not, or when it was last updated, there is no updated data to give for now, so I'm fine with adding that later

reivilibre: Yes, also sending the password hash back to the user doesn't seem like a desirable thing anyway ;-)

crates/handlers/src/graphql/mutations/user.rs
reivilibre: The only counterpoint I can think of is if you were a server administrator (who can change their own password without their current password. Not sure if this is dodgy or not.) but then your administrator privilege is taken away before you try to change your password that way.

However you could say the same thing about editing another user's password. So yes, let's just throw the big error either way I guess.

sandhose (approved):

PR #2863, opened 2024-06-20
Add Self-service Password Change

### LINKED ISSUE #2148: Better UI for self-served password change
When an account is baked by a password credential, the end user should be able to change their password themselves. This assumes the user have not lost their password, which is different from the password recovery flow covered in #13 

The flow would look like this:

 - the user go to their MAS user management screen
 - they get asked for their old password, their new password and a confirmation of the new password
 - their other sessions get (optionally?) all revoked
 - (optionally) they get an email to warn them their password change

Subtasks:

 - #2197 

Open questions:

 - What knobs should we expose to server admins?
   - Allow password change overall
   - Notify of password changes by email
   - Whether other sessions should be signed out
 - Is the password change email notification useful? If so, we need to set the wording/design
 - Should we actually revoke sessions? Maybe a middle ground would be to revoke *browsers sessions*, so sessions in the MAS UI, but *not* client sessions? That would prevent potential attackers to change sensitive stuff on the account, e.g. the list of emails, delete the account, etc.

Relevant design screens:

 - [Potential entry point for this flow, the account mgmt screen](https://www.figma.com/file/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?type=design&node-id=4344-10620&mode=design&t=fPuEstdLnvQOOyBL-4)

reivilibre:
This is still missing some polish, like:

- #2855
- #2854
- #2862

Fixes #2148.


frontend/src/routes/password.change.tsx
```diff
@@ -0,0 +1,293 @@
...
+    result.data && result.data.setPassword.status == SetPasswordStatus.Allowed;
+  const handleableError =
+    result.data && result.data.setPassword.status != SetPasswordStatus.Allowed;
+  const unhandleableError = result.error !== undefined;
+
+  const errorMsg: string | undefined = ((): string | undefined => {
+    switch (result.data?.setPassword.status) {
+      case SetPasswordStatus.InvalidNewPassword:
+        return t(
+          "frontend.password_change.failure.description_INVALID_NEW_PASSWORD",
```
sandhose, line 125: That would be the first time we'd use ALL CAPS in a translation key. I'd rather nest, like
```suggestion
          "frontend.password_change.failure.description.invalid_new_password",
```

frontend/src/routes/password.change.tsx
```diff
@@ -0,0 +1,293 @@
...
+            which could be logged, if for some reason the event handler fails.
+          */}
+
+          {!success && (
+            <>
+              <Form.Field
+                name="current_password"
+                serverInvalid={
+                  result.data &&
+                  result.data.setPassword.status ==
```
sandhose, line 168: You could shorten this with an optional chaining: 
```suggestion
                  result.data?.setPassword.status ===
```

(also triple equals please :))

frontend/src/routes/password.change.tsx
```diff
@@ -0,0 +1,293 @@
...
+
+  const success =
+    result.data && result.data.setPassword.status == SetPasswordStatus.Allowed;
+  const handleableError =
+    result.data && result.data.setPassword.status != SetPasswordStatus.Allowed;
+  const unhandleableError = result.error !== undefined;
+
+  const errorMsg: string | undefined = ((): string | undefined => {
+    switch (result.data?.setPassword.status) {
+      case SetPasswordStatus.InvalidNewPassword:
```
sandhose, line 123: Could this one be inline on the `new_password` field?

frontend/src/routes/password.change.tsx
```diff
@@ -0,0 +1,293 @@
...
+          "frontend.password_change.failure.description_NO_CURRENT_PASSWORD",
+        );
+      case SetPasswordStatus.NotAllowed:
+        return t("frontend.password_change.failure.description_NOT_ALLOWED");
+      case SetPasswordStatus.NotFound:
+        return t("frontend.password_change.failure.description_NOT_FOUND");
+      case SetPasswordStatus.PasswordChangesDisabled:
+        return t(
+          "frontend.password_change.failure.description_PASSWORD_CHANGES_DISABLED",
+        );
```
sandhose, line 138: Actually for those I would just `throw` within the component? None of those are not recoverable by the user, so showing a page 'crash' is probably the best thing?

frontend/src/routes/password.change.tsx
```diff
@@ -0,0 +1,293 @@
...
+        />
+
+        <Form.Root onSubmit={onSubmit} method="POST">
+          {/*
+            In normal operation, the submit event should be `preventDefault()`ed.
+            method = POST just prevents sending passwords in the query string,
+            which could be logged, if for some reason the event handler fails.
+          */}
+
+          {!success && (
```
sandhose, line 162: What about redirecting to a `success` page on success instead? So that we already have the redirect logic, and can do later on the toast

frontend/src/routes/password.change.tsx
```diff
@@ -0,0 +1,293 @@
...
+              type="critical"
+              title={t("frontend.password_change.failure.title")}
+            >
+              {errorMsg}
+            </Alert>
+          )}
+
+          <ButtonLink to="/" kind="tertiary">
+            {(result.data && t("action.back")) || t("action.cancel")}
+          </ButtonLink>
```
sandhose, line 288: Making a dedicated success page would avoid having this kind of logic within the component

frontend/src/routes/password.change.tsx
```diff
@@ -0,0 +1,293 @@
...
+              <Form.Field name="new_password">
+                <Form.Label>
+                  {t("frontend.password_change.new_password_label")}
+                </Form.Label>
+
+                <Form.PasswordControl
+                  required
+                  autoComplete="new-password"
+                  ref={newPasswordRef}
+                  onBlur={() => newPasswordAgainRef.current!.reportValidity()}
```
sandhose, line 208: I don't really like that it triggers on the first run. Maybe check whether the value is non empty? 
```suggestion
                  onBlur={() => newPasswordAgainRef.current!.value && newPasswordAgainRef.current!.reportValidity()}
```

sandhose (changes requested):
A few things here and there, but that looks really nice!

frontend/src/routes/password.change.tsx
reivilibre: some of these I'm on the fence about.


e.g. password change disabled: the functionality could have been disabled since you navigated to the page, it feels a bit crap to show a generic 'something went wrong' error when we know the actual reason.

not allowed / not found: these probably make more sense if we had an admin control panel in MAS, but agreed we can probably lose them

no current password: I'd be tempted to keep it, since it's explanatory of a real-world error you might hit?

frontend/src/routes/password.change.tsx
reivilibre: orginally I was building this dynamically from the content of the enum, but that upset quite a few assumptions that we only use static translation keys, so that's why this came about ;-)

frontend/src/routes/_account.index.tsx
```diff
@@ -109,6 +112,14 @@ function Index(): React.ReactElement {
 
         <Separator />
 
+        {siteConfig.passwordLoginEnabled && (
+          <div className="flex flex-col gap-4" id="password">
+            <AccountManagementPasswordPreview siteConfig={siteConfig} />
+          </div>
```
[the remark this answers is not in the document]
reivilibre: I stole this from somewhere so I'd have to experiment to remove it

frontend/src/components/AccountManagementPasswordPreview/AccountManagementPasswordPreview.module.css
```diff
@@ -0,0 +1,36 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+.password-preview-field {
+  flex: 1;
+}
+
+.link {
```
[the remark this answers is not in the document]
reivilibre: the `Link` component does not have the underline or the white text colour, it just looks like plain grey text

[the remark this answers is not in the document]
sandhose: Fundamentally we're missing a good implementation of the Text Link component: https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=645-3494&t=OH2MjoxKqsJTB1vF-4

I'll leave the duplication for now, and will see to implement that in Compound

frontend/src/routes/password.change.tsx
```diff
@@ -0,0 +1,283 @@
...
+          </Form.Submit>
+
+          {handleableError && (
+            <Alert
+              type="critical"
+              title={t("frontend.password_change.failure.title")}
+            >
+              {errorMsg}
+            </Alert>
+          )}
```
sandhose, line 274: I'm not sure I like the `switch/case` logic up there. Because we have some errors specific to fields, I would rather have a series of `{result.data?.setPassword.status === SetPasswordStatus.NoCurrentPassword && (<Alert … />)}`. wdyt?

Also could you stick alerts at the top of the form?

frontend/src/routes/password.change.tsx
```diff
@@ -0,0 +1,283 @@
...
+    const newPasswordAgain = formData.get("new_password_again") as string;
+
+    if (newPassword !== newPasswordAgain) {
+      throw new Error("passwords mismatch; this should be checked by the form");
+    }
+
+    const response = await changePassword({ userId, oldPassword, newPassword });
+
+    if (response.data?.setPassword.status === SetPasswordStatus.Allowed) {
+      router.navigate({ to: "/password/change_success" });
```
sandhose, line 104: I'm nitpicking at this point, but I personally don't like underscores in paths, and we used dashes elsewhere. Can you make this either `/password/change/success` or `/password/change-success`?

sandhose (changes requested):
A few nitpicks but we're almost there!

frontend/src/routes/password.change.tsx
reivilibre: heh I am the opposite, I don't like the dashes in paths.

But anyway, I tried `/` but I must have been doing something wrong — it just renders the change password form instead of the success page so I gave up on that. Will change to dashes

sandhose: that's because of how the router nests components

`password.change.success.tsx` will basically render `Password > Change > Success`, nesting components using the `<Outlet />` thingy.

You'd need to move the `password.change.tsx` page to `password.change.index.tsx`

see: https://tanstack.com/router/latest/docs/framework/react/guide/route-trees#flat-routes

reivilibre: aaah. That makes more sense, but still grrr :-). I will try that

reivilibre: yeah that appears to work

frontend/src/routes/password.change_success.tsx
```diff
@@ -0,0 +1,81 @@
...
+          subtitle={t("frontend.password_change.subtitle")}
+          success
+        />
+
+        <Alert
+          type="success"
+          title={t("frontend.password_change.success.title")}
+        >
+          {t("frontend.password_change.success.description")}
+        </Alert>
```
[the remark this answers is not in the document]
reivilibre: think this should be more what you had in mind, let me know

frontend/src/routes/password.change_success.tsx
```diff
@@ -0,0 +1,81 @@
...
+const CURRENT_VIEWER_QUERY = graphql(/* GraphQL */ `
+  query CurrentViewerQuery {
+    viewer {
+      __typename
+      ... on Node {
+        id
+      }
+    }
+  }
+`);
```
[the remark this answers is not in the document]
reivilibre: yeah I was on the fence about this one...

frontend/src/routes/password.change.tsx
reivilibre: do you think we should not show the errors in an alert if they are specific to a form field? I was thinking that might make it a bit harder to notice but wasn't entirely sure.

Both the switch and the sequence of conditionally rendered alerts feel a bit tedious but struggling to think of anything better

reivilibre: that said: I think it's important to have a 'default' case (for unhandled errors etc) and by taking your approach, it's harder to see how we'd do that

sandhose: > do you think we should not show the errors in an alert if they are specific to a form field? I was thinking that might make it a bit harder to notice but wasn't entirely sure.

Yep, as radix forms will take care of focusing the errored field. I don't think I've saw any form with an alert like that in the designs

> that said: I think it's important to have a 'default' case (for unhandled errors etc) and by taking your approach, it's harder to see how we'd do that

that's a fair point. Although could be offset by doing something like:

```typescript
  const status = result.data?.setPassword.status;
  switch (status) {
    // Cases that are handled by the form
    case SetPasswordStatus.InvalidNewPassword:
    case SetPasswordStatus.NoCurrentPassword:
    case SetPasswordStatus.PasswordChangesDisabled:
    case SetPasswordStatus.WrongPassword:
    case SetPasswordStatus.Allowed:
    case undefined:
      break;

    case SetPasswordStatus.NotFound:
    case SetPasswordStatus.NotAllowed:
      throw new Error(`unexpected error when changing password: ${status}`);

    default:
      unreachable(status);
  }
```

sandhose (approved):
I'm fine with the double-error message for now, will see during design review I guess wether we should change that or not

Issue #2785, opened 2024-05-17
Rewrite the async job system


sandhose:
We're currently using apalis for async jobs, which… is probably not mature enough.
A few issues with it:

 - the crates are a bit messy
 - they are doing breaking changes all the time. See #2784 
 - the database structure is… weird
 - the jobs don't retry correctly
 - the "workers" crash when it looses the connection and don't restart
 - sharing state in the job is not type safe

matrixbot:
For your information, this issue has been copied over to the Element fork of `matrix-authentication-service`: <https://github.com/element-hq/matrix-authentication-service/issues/2785>

PR #2866, opened 2024-06-21
Forgot password flow

### LINKED ISSUE #13: Password recovery
Users should be able to recover their account via email.

Potential flows:

 1. Start the recovery, you get a code by email, you enter that code, you can set a new password
 2. Start the recovery, you get a link by email, you follow that link, you can set a new password

The first flow feels better at not disrupting the current action. If you're in the middle of a client login, it's easier to resume that login after that.

The second flow feels better at preventing social engineering attacks, as we would require the person to click a link and change the password on the same device they are checking their emails, whereas in the option 1., the attacker could just ask "can you give me the code you just got by email" and the user could overlook that it's for a password change?

Open questions:

 - Option 1. or 2. -> 2
 - What's the input for resetting the account? The Matrix ID or the email address? -> email only
 - Email design and wording
 - Do we automatically login after password reset? -> no
 - Do we send a password change email notification after that? -> ideally
 - Do we log off existing sessions when you recover your account that way? -> no
 - Do we only use the primary email address, or do we allow *any* email address on the account? -> all emails, but add email notification if we do so

Relevant design screens:

 - [Forgot password flow](https://www.figma.com/file/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?type=design&node-id=3784-2826&mode=design&t=fPuEstdLnvQOOyBL-4)

sandhose:
Fixes #13 

This implements the password recovery flow as designed [here](https://www.figma.com/design/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?node-id=5354-4833&t=JASaGhnRh6Gj6FU2-4).

The feature is disabled by default.

A few things to do still to exactly match the designs:

 - The password form isn't interactive at all
 - The recovery email is missing the ['user box'](https://www.figma.com/design/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?node-id=5352-10257&t=JASaGhnRh6Gj6FU2-4). This is mostly because querying the displayname was potentially annoying here, and I'm not sure about implementing that box in an email (`<table>` maybe?)
 - We're not sending a confirmation email once the password has been changed. I think we should do this separately to also add the same notification to the password change flow

Open questions:

 - Should the initial form have a CAPTCHA? It's basically an open form to spam someone's email
 - I wonder if we should move that page to the React app? That would work by extending the `setPassword` API to accept a recovery 'ticket', and make it easier to have the same form validation logic

I made it so that it's reviewable commit by commit

### DIFF
--- a/crates/data-model/src/users.rs
+++ b/crates/data-model/src/users.rs
@@ -1,4 +1,4 @@
-// Copyright 2021 The Matrix.org Foundation C.I.C.
+// Copyright 2021-2024 The Matrix.org Foundation C.I.C.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -77,6 +77,44 @@ pub enum AuthenticationMethod {
     Password { user_password_id: Ulid },
     UpstreamOAuth2 { upstream_oauth2_session_id: Ulid },
     Unknown,
+}
+
+/// A session to recover a user if they have lost their credentials
+///
+/// For each session intiated, there may be multiple [`UserRecoveryTicket`]s
+/// sent to the user, either because multiple [`User`] have the same email
+/// address, or because the user asked to send the recovery email again.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct UserRecoverySession {
+    pub id: Ulid,
+    pub email: String,
+    pub user_agent: UserAgent,
+    pub ip_address: Option<IpAddr>,
+    pub locale: String,
+    pub created_at: DateTime<Utc>,
+    pub consumed_at: Option<DateTime<Utc>>,
+}
+
+/// A single recovery ticket for a user recovery session
+///
+/// Whenever a new recovery session is initiated, a new ticket is created for
+/// each email address matching in the database. That ticket is sent by email,
+/// as a link that the user can click to recover their account.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize)]
+pub struct UserRecoveryTicket {
+    pub id: Ulid,
+    pub user_recovery_session_id: Ulid,
+    pub user_email_id: Ulid,
+    pub ticket: String,
+    pub created_at: DateTime<Utc>,
+    pub expires_at: DateTime<Utc>,
+}
+
+impl UserRecoveryTicket {
+    #[must_use]
+    pub fn active(&self, now: DateTime<Utc>) -> bool {
+        now < self.expires_at
+    }
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize)]
--- a/crates/handlers/src/views/recovery/finish.rs
+++ b/crates/handlers/src/views/recovery/finish.rs
@@ -1,0 +1,246 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use anyhow::Context;
+use axum::{
+    extract::{Query, State},
+    response::{Html, IntoResponse, Response},
+    Form,
+};
+use mas_axum_utils::{
+    cookies::CookieJar,
+    csrf::{CsrfExt, ProtectedForm},
+    FancyError,
+};
+use mas_policy::Policy;
+use mas_router::UrlBuilder;
+use mas_storage::{BoxClock, BoxRepository, BoxRng};
+use mas_templates::{
+    ErrorContext, FieldError, FormState, RecoveryFinishContext, RecoveryFinishFormField,
+    TemplateContext, Templates,
+};
+use serde::{Deserialize, Serialize};
+use zeroize::Zeroizing;
+
+use crate::{passwords::PasswordManager, PreferredLanguage};
+
+#[derive(Deserialize)]
+pub(crate) struct RouteQuery {
+    ticket: String,
+}
+
+#[derive(Deserialize, Serialize)]
+pub(crate) struct RouteForm {
+    new_password: String,
+    new_password_confirm: String,
+}
+
+pub(crate) async fn get(
+    mut rng: BoxRng,
+    clock: BoxClock,
+    mut repo: BoxRepository,
+    State(templates): State<Templates>,
+    PreferredLanguage(locale): PreferredLanguage,
+    cookie_jar: CookieJar,
+    Query(query): Query<RouteQuery>,
+) -> Result<Response, FancyError> {
+    let (csrf_token, cookie_jar) = cookie_jar.csrf_token(&clock, &mut rng);
+
+    let ticket = repo
+        .user_recovery()
+        .find_ticket(&query.ticket)
+        .await?
+        .context("Unknown ticket")?;
+
+    let session = repo
+        .user_recovery()
+        .lookup_session(ticket.user_recovery_session_id)
+        .await?
+        .context("Unknown session")?;
+
+    if !ticket.active(clock.now()) || session.consumed_at.is_some() {
+        // TODO: render a 'link expired' page
+        let rendered = templates.render_error(
+            &ErrorContext::new()
+                .with_code("Link expired")
+                .with_language(&locale),
+        )?;
+        return Ok((cookie_jar, Html(rendered)).into_response());
+    }
+
+    let user_email = repo
+        .user_email()
+        .lookup(ticket.user_email_id)
+        .await?
+        // Only allow confirmed email addresses
+        .filter(|email| email.confirmed_at.is_some())
+        .context("Unknown email address")?;
+
+    let user = repo
+        .user()
+        .lookup(user_email.user_id)
+        .await?
+        .context("Invalid user")?;
+
+    if !user.is_valid() {
+        // TODO: render a 'account locked' page
+        let rendered = templates.render_error(
+            &ErrorContext::new()
+                .with_code("Account locked")
+                .with_language(&locale),
+        )?;
+        return Ok((cookie_jar, Html(rendered)).into_response());
+    }
+
+    let context = RecoveryFinishContext::new(user)
+        .with_csrf(csrf_token.form_value())
+        .with_language(locale);
+
+    let rendered = templates.render_recovery_finish(&context)?;
+
+    Ok((cookie_jar, Html(rendered)).into_response())
+}
+
+pub(crate) async fn post(
+    mut rng: BoxRng,
+    clock: BoxClock,
+    mut repo: BoxRepository,
+    mut policy: Policy,
+    State(password_manager): State<PasswordManager>,
+    State(templates): State<Templates>,
+    State(url_builder): State<UrlBuilder>,
+    PreferredLanguage(locale): PreferredLanguage,
+    cookie_jar: CookieJar,
+    Query(query): Query<RouteQuery>,
+    Form(form): Form<ProtectedForm<RouteForm>>,
+) -> Result<Response, FancyError> {
+    let (csrf_token, cookie_jar) = cookie_jar.csrf_token(&clock, &mut rng);
+
+    let ticket = repo
+        .user_recovery()
+        .find_ticket(&query.ticket)
+        .await?
+        .context("Unknown ticket")?;
+
+    let session = repo
+        .user_recovery()
+        .lookup_session(ticket.user_recovery_session_id)
+        .await?
+        .context("Unknown session")?;
+
+    if !ticket.active(clock.now()) || session.consumed_at.is_some() {
+        // TODO: render a 'link expired' page
+        let rendered = templates.render_error(
+            &ErrorContext::new()
+                .with_code("Link expired")
+                .with_language(&locale),
+        )?;
+        return Ok((cookie_jar, Html(rendered)).into_response());
+    }
+
+    let user_email = repo
+        .user_email()
+        .lookup(ticket.user_email_id)
+        .await?
+        // Only allow confirmed email addresses
+        .filter(|email| email.confirmed_at.is_some())
+        .context("Unknown email address")?;
+
+    let user = repo
+        .user()
+        .lookup(user_email.user_id)
+        .await?
+        .context("Invalid user")?;
+
+    if !user.is_valid() {
+        // TODO: render a 'account locked' page
+        let rendered = templates.render_error(
+            &ErrorContext::new()
+                .with_code("Account locked")
+                .with_language(&locale),
+        )?;
+        return Ok((cookie_jar, Html(rendered)).into_response());
+    }
+
+    let form = cookie_jar.verify_form(&clock, form)?;
+
+    // Check the form
+    let mut form_state = FormState::from_form(&form);
+
+    if form.new_password.is_empty() {
+        form_state = form_state
+            .with_error_on_field(RecoveryFinishFormField::NewPassword, FieldError::Required);
+    }
+
+    if form.new_password_confirm.is_empty() {
+        form_state = form_state.with_error_on_field(
+            RecoveryFinishFormField::NewPasswordConfirm,
+            FieldError::Required,
+        );
+    }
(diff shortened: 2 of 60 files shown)


crates/handlers/src/views/recovery/start.rs
```diff
@@ -0,0 +1,137 @@
...
+    if maybe_session.is_some() {
+        // TODO: redirect to continue whatever action was going on
+        return Ok((cookie_jar, url_builder.redirect(&mas_router::Index)).into_response());
+    }
+
+    let context = RecoveryStartContext::new()
+        .with_csrf(csrf_token.form_value())
+        .with_language(locale);
+
+    repo.save().await?;
```
reivilibre, line 76: why save?

crates/handlers/src/views/recovery/start.rs
```diff
@@ -0,0 +1,137 @@
...
+    let form = cookie_jar.verify_form(&clock, form)?;
+    let mut form_state = FormState::from_form(&form);
+
+    if Address::from_str(&form.email).is_err() {
+        form_state =
+            form_state.with_error_on_field(RecoveryStartFormField::Email, FieldError::Invalid);
+    }
+
+    if !form_state.is_valid() {
+        repo.save().await?;
```
reivilibre, line 123: also why save here

templates/pages/recovery/start.html
```diff
@@ -0,0 +1,48 @@
...
+    </div>
+
+    <div class="header">
+      <h1 class="title">{{ _("mas.recovery.start.heading") }}</h1>
+      <p class="text">{{ _("mas.recovery.start.description") }}</p>
+    </div>
+  </header>
+
+  <form class="cpd-form-root" method="POST">
+    {% if form.errors is not empty %}
```
reivilibre, line 32: meh, but this if condition seems to be redundant

translations/en.json
```diff
@@ -356,6 +356,20 @@
...
       }
     },
+    "recovery": {
+      "start": {
+        "description": "An email will be sent with a link to reset your password.",
+        "@description": {
+          "context": "pages/recovery/start.html:27:25-60",
+          "description": "The description of the page to initiate an account recovery"
+        },
+        "heading": "Enter your email to continue",
```
reivilibre, line 492: should this say email *address*?

crates/storage/src/job.rs
```diff
@@ -432,8 +432,40 @@ mod jobs {
     impl Job for DeactivateUserJob {
         const NAME: &'static str = "deactivate-user";
     }
+
+    /// Send account recovery emails
+    #[derive(Serialize, Deserialize, Debug, Clone)]
+    pub struct SendAccountRecoveryEmailJob {
```
reivilibre, line 438: maybe make this plural (Email**s**) so it's a bit clearer it does multiple of them

crates/tasks/src/recovery.rs
```diff
@@ -0,0 +1,105 @@
...
+
+pub(crate) fn register(
+    suffix: &str,
+    monitor: Monitor<TokioExecutor>,
+    state: &State,
+    storage_factory: &PostgresStorageFactory,
+) -> Monitor<TokioExecutor> {
+    let send_user_recovery_email_worker = crate::build!(SendAccountRecoveryEmailJob => send_account_recovery_email_job, suffix, state, storage_factory);
+
+    monitor.register(send_user_recovery_email_worker)
```
reivilibre, line 140: this seems like cool witchcraft, I will have to look at how this works someda

crates/tasks/src/recovery.rs
```diff
@@ -0,0 +1,105 @@
...
+        for email in page.edges {
+            let ticket = Alphanumeric.sample_string(&mut rng, 32);
+
+            let _ticket = repo
+                .user_recovery()
+                .add_ticket(&mut rng, &clock, &session, &email, ticket)
+                .await?;
+
+            info!("Sending recovery email to {}", email.email);
+            // TODO
```
reivilibre: I guess fair enough that this is left until later, but it'd be good to make sure this is robust e.g. one e-mail failing shouldn't prevent another from being tried

translations/en.json
```diff
@@ -383,6 +383,28 @@
       }
     },
     "recovery": {
+      "progress": {
+        "change_email": "Try a different email",
+        "@change_email": {
+          "context": "pages/recovery/progress.html:38:33-72",
+          "description": "Button to change the email address for the password recovery link"
+        },
+        "description": "We send an email with a link to reset your password if there's an account using <span>%(email)s</span>.",
```
reivilibre, line 470: sen**t**

translations/en.json
```diff
@@ -383,6 +383,28 @@
       }
     },
     "recovery": {
+      "progress": {
+        "change_email": "Try a different email",
```
reivilibre, line 465: email *address*?

crates/handlers/src/views/recovery/progress.rs
```diff
@@ -62,3 +70,51 @@ pub(crate) async fn get(
...
+        return Ok((cookie_jar, url_builder.redirect(&mas_router::Index)).into_response());
+    }
+
+    let Some(recovery_session) = repo.user_recovery().lookup_session(id).await? else {
+        // XXX: is that the right thing to do?
+        return Ok((
+            cookie_jar,
+            url_builder.redirect(&mas_router::AccountRecoveryStart),
+        )
+            .into_response());
```
reivilibre, line 121: maybe we should say the link 'expired' or something. I guess.

crates/handlers/src/views/recovery/finish.rs
```diff
@@ -0,0 +1,246 @@
...
+        .user_recovery()
+        .find_ticket(&query.ticket)
+        .await?
+        .context("Unknown ticket")?;
+
+    let session = repo
+        .user_recovery()
+        .lookup_session(ticket.user_recovery_session_id)
+        .await?
+        .context("Unknown session")?;
```
reivilibre, line 78: This should probably say something like it was invalid or expired etc, rather than 500ing

crates/handlers/src/views/recovery/finish.rs
```diff
@@ -0,0 +1,246 @@
...
+
+    if !user.is_valid() {
+        // TODO: render a 'account locked' page
+        let rendered = templates.render_error(
+            &ErrorContext::new()
+                .with_code("Account locked")
+                .with_language(&locale),
+        )?;
+        return Ok((cookie_jar, Html(rendered)).into_response());
+    }
```
reivilibre, line 198: all the logic above this seems to be in common with the `get` route; wonder if can we pull this out into a separate function without it being too awkward?

reivilibre (changes requested):
seems reasonable to me overall. I'll spin it up in a browser tomorrow but ran out of time today

crates/handlers/src/views/recovery/finish.rs
sandhose: Later on I added the 'consumed at' check separately, so that exact error would only ever happen if the foreign key is broken for some reason?

crates/handlers/src/views/recovery/progress.rs
sandhose: I mean it's really that you got to the wrong place, as `lookup_session` will give you even consumed/expired sessions. A 404 would be better, but we don't have a nice way to display that page.

I think we should track in a separate issue to have nice 404 pages for when you go to pages with an unknown ID, because we have many cases like that

crates/handlers/src/views/recovery/start.rs
sandhose: I did read somewhere that `commit`ing a transaction, even if it was read only, was better than rolling it back? (which is effectively what `Drop`ping the repository will do). Although you're right that this isn't consistant with anything. Should I add `save`s everywhere or drop this one?

crates/handlers/src/views/recovery/start.rs
sandhose: same reason

crates/storage-pg/migrations/20240621080509_user_recovery.sql
```diff
@@ -0,0 +1,64 @@
...
+  -- The email address for which the recovery session was requested
+  "email" TEXT NOT NULL,
+
+  -- The user agent of the client that requested the recovery session
+  "user_agent" TEXT NOT NULL,
+
+  -- The IP address of the client that requested the recovery session
+  "ip_address" INET,
+
+  -- The languag eof the client that requested the recovery session
```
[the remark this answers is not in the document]
sandhose: Done in bccd433c3b8a835ebd855f39ad681c6df5bce13f

crates/storage-pg/src/user/recovery.rs
```diff
@@ -0,0 +1,337 @@
...
+
+#[async_trait]
+impl<'c> UserRecoveryRepository for PgUserRecoveryRepository<'c> {
+    type Error = DatabaseError;
+
+    #[tracing::instrument(
+        name = "db.user_recovery.lookup_session",
+        skip_all,
+        fields(
+            db.statement,
```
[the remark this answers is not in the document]
sandhose: There is a `traced` method on the `ExecuteExt` trait which will fill the `db.statement` on the current span. So this adds an empty `db.statement` field, and later on, before executing the query, we call `.traced()` on it to record the SQL associated to it

crates/storage-pg/src/user/recovery.rs
```diff
@@ -0,0 +1,337 @@
...
+    async fn consume_ticket(
+        &mut self,
+        clock: &dyn Clock,
+        user_recovery_ticket: UserRecoveryTicket,
+        mut user_recovery_session: UserRecoverySession,
+    ) -> Result<UserRecoverySession, Self::Error> {
+        // We don't really use the ticket, we just want to make sure we drop it
+        let _ = user_recovery_ticket;
+
+        // This should have been checked by the caller
```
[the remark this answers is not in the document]
sandhose: Done in  fae7ce184cd11312f1558fb6c4ed69f62c225f74

crates/storage/src/job.rs
sandhose: Done in 369d8b1b

crates/tasks/src/recovery.rs
sandhose: errmmm you probably don't know to know about this :D 

#2785 😇

translations/en.json
sandhose: I just copied what's in the designs: https://www.figma.com/design/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?node-id=5354-3370&t=lkEDnlQD00OwKSmK-4

translations/en.json
sandhose: I just copied what's in the designs: https://www.figma.com/design/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?node-id=5354-3370&t=lkEDnlQD00OwKSmK-4

translations/en.json
sandhose: Done in c401045742f8821567a4b63bd2746a547185103d

crates/tasks/src/recovery.rs
sandhose: That was done later in b51fbe4b

crates/data-model/src/users.rs
```diff
@@ -79,6 +79,41 @@ pub enum AuthenticationMethod {
     Unknown,
 }
 
+/// A session to recover a user if they have lost their credentials
+///
+/// This can trigger send an email to many 'users' at the same time, which is
+/// why it has a one-to-many relationship with [`UserRecoveryTicket`].
```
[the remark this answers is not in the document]
reivilibre: huh, surprised multiple users can have the same e-mail address but if we're sure that's what we want then that makes more sense, yes!

crates/storage-pg/src/user/recovery.rs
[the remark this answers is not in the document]
reivilibre: aaaah OK

crates/handlers/src/views/recovery/progress.rs
sandhose: https://github.com/matrix-org/matrix-authentication-service/issues/2880

crates/handlers/src/views/recovery/start.rs
reivilibre: that was sort of part of my point yes, we have some early returns where we don't save.

There is a thing in Postgres, at least under the SERIALIZABLE isolation level, where when you read something, you can't trust it until you finish your transaction, commit and it commits successfully. I thought I wrote a blog post explaining the mechanics of this but it seems I don't have it. Maybe it's unpublished on a different computer :(.

The high-level summary from Postgres' manual is:

> When relying on Serializable transactions to prevent anomalies, it is important that any data read from a permanent user table not be considered valid until the transaction which read it has successfully committed. This is true even for read-only transactions, except that data read within a deferrable read-only transaction is known to be valid as soon as it is read, because such a transaction waits until it can acquire a snapshot guaranteed to be free from such problems before starting to read any data. In all other cases applications must not depend on results read during a transaction that later aborted; instead, they should retry the transaction until it succeeds.
> 
> — https://www.postgresql.org/docs/current/transaction-iso.html

Annoyingly I don't think MAS is well-structured for this. (I'm not sure how you can do this ergonomically without having to structure your DB code like Synapse does: where you pass a transaction function and Synapse retries it as needed.)
But I'm not sure you use the serialisable transaction isolation level anyway and I can't remember if repeatable read has the same caveat for read-only transactions.


I am not sure which side of the fence to sit on here, but I don't think it's easy to remember to save before every early exit. :/

translations/en.json
reivilibre: fair enough. It seems a bit casual to me but maybe that's what we're going for.

crates/tasks/src/recovery.rs
reivilibre: yup so it was

crates/handlers/src/views/recovery/progress.rs
reivilibre: well this kind of raises another Q to me: Do we really want to keep old expired sessions around in the DB forever?

I can't say I see much reason to keep them for a very long time, because it's just going to be ever-growing cruft.

crates/handlers/src/views/recovery/finish.rs
reivilibre: I think I mean for the 'Unknown ticket' in this case then

crates/handlers/src/views/recovery/finish.rs
reivilibre: any insight why this is hard? what happens if you try it?

crates/handlers/src/views/recovery/progress.rs
sandhose: Generally we don't cleanup most things in the database. I rather soft-delete stuff and keep things for now, and introduce maintenance/cleanup tasks later once we have a nice way of doing this?

crates/handlers/src/views/recovery/finish.rs
sandhose: That wouldn't be the first place we have this. As we're not deleting tickets for now, it's probably fine, and then it's a case of #2880?

crates/data-model/src/users.rs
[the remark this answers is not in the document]
sandhose: Tracked as https://github.com/matrix-org/matrix-authentication-service/issues/2881

crates/handlers/src/views/recovery/start.rs
sandhose: We're indeed not using the SERIALIZABLE isolation level. Most of the time if we do an early exit, it's because we're in an error state anyway, so it probably doesn't matter? Not sure what to do here

templates/pages/recovery/start.html
sandhose: we have this pattern in many places, at least it's consistent 🙈

crates/handlers/src/views/recovery/progress.rs
reivilibre: cynical take: that's how you get Synapse ;-)

crates/handlers/src/views/recovery/start.rs
reivilibre: yeah. Maybe we just leave it as is. It's not super satisfying but not something to get stuck on too much maybe

crates/handlers/src/views/recovery/finish.rs
reivilibre: ok for now

reivilibre (approved):

Issue #172, opened 2022-05-04
Enforcement of password strength


hughns:

sandhose:
There is some password strength enforcement via the OPA policy + config (require uppercase/lowercase/number + min length), although we might be better off using something more intelligent like [zxcvbn](https://github.com/dropbox/zxcvbn)? This one would definitely require interactive feedback on the frontend, else you'd get cryptic error messages like "your password is not strong enough" without knowing exactly what "strong enough" means

americanrefugee:
Here is [the final component in Figma](https://www.figma.com/design/rTaQE2nIUSLav4Tg3nozq7/Compound-Web-Components?node-id=5425-39), and here is a [reference in the designs](https://www.figma.com/design/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?node-id=5315-21631&t=8c158B7ZK0TTWzni-4).

reivilibre:
After #2972, here's what will be left:

- [ ] converting the other password screens to use the React frontend so they can give the client-side feedback

matrixbot:
For your information, this issue has been copied over to the Element fork of `matrix-authentication-service`: <https://github.com/element-hq/matrix-authentication-service/issues/172>

PR #2875, opened 2024-06-26
Remove OPA-based password policy enforcement

reivilibre:
The OPA-based policies are too opaque and complicated; this is a precursor to just using the zxcvbn algorithm, which should let us provide better feedback in real-time to the users. (#172)


sandhose (approved):
I merged back main as there was a conflict. Sorry that I forgot about this PR, LGTM now 👍

PR #2885, opened 2024-06-28
Make the b64decode filter try multiple base64 encoding variants

sandhose:
This was sometimes a problem when decoding the subject given by Dex

### DIFF
--- a/crates/handlers/src/upstream_oauth2/template.rs
+++ b/crates/handlers/src/upstream_oauth2/template.rs
@@ -14,7 +14,7 @@
 
 use std::{collections::HashMap, sync::Arc};
 
-use base64ct::{Base64, Encoding};
+use base64ct::{Base64, Base64Unpadded, Base64Url, Base64UrlUnpadded, Encoding};
 use minijinja::{Environment, Error, ErrorKind, Value};
 
 fn split(value: &str, separator: Option<&str>) -> Vec<String> {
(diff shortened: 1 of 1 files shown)


crates/handlers/src/upstream_oauth2/template.rs
```diff
@@ -25,13 +25,19 @@ fn split(value: &str, separator: Option<&str>) -> Vec<String> {
...
 
 fn b64decode(value: &str) -> Result<Value, Error> {
-    let bytes = Base64::decode_vec(value).map_err(|e| {
-        Error::new(
-            ErrorKind::InvalidOperation,
-            "Failed to decode base64 string",
-        )
-        .with_source(e)
-    })?;
+    // We're not too concerned about the performance of this filter, so we'll just
```
reivilibre, line 28: This feels a bit lax TBH, I'm not sure about the intended usage but this is the kind of thing that makes you accidentally accept something you didn't mean to and then make it hard to break that later.

If we're really sure this is fine then fine, but otherwise I'd prefer to have different filters for the different variants

reivilibre (approved):

crates/handlers/src/upstream_oauth2/template.rs
sandhose: This *can* be used when transforming attributes from an upstream IDP. We use this for decoding the Dex subject in ESS, and as we're not documenting this, I think it'll be fine if we want to break it later

PR #2886, opened 2024-06-28
Pull out common code from both `get` and `post` of the recovery handler

reivilibre:
From https://github.com/matrix-org/matrix-authentication-service/pull/2866#discussion_r1658701097
It was bugging me to find out why we couldn't do this ... I wanted to witness the errors for myself and think about why it isn't Send+Sync etc.

Then it turned out it's fine.

It's a bit verbose (it's actually more code) but it does mean that we won't accidentally forget to update both checks at the same time, so I am mildly in favour, but happy to hear what you think

### DIFF
--- a/crates/handlers/src/views/recovery/finish.rs
+++ b/crates/handlers/src/views/recovery/finish.rs
@@ -20,10 +20,11 @@ use axum::{
 };
 use mas_axum_utils::{
     cookies::CookieJar,
-    csrf::{CsrfExt, ProtectedForm},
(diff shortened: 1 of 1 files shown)


sandhose (commented):
I'm really not sure about this pattern. I wonder if instead, we should have a common handler, and extract the `http::Method` and make the form `Option`al, and diverge depending on those? And then in the router `get(views::recovery::finish::handler).post(views::recovery::finish::handler)`?

reivilibre:
didn't consider that, that could work

PR #2908, opened 2024-06-28
fix class-b RFC1918 network

ctr49:
I'm pretty sure this should be 192.168.0.0/16, not 192.128.0.0/16 as per RFC1918. Nothing against AT&T (who own 192.128.0.0/16)...


sandhose:
Good catch! Indeed, nothing against AT&T :)

This file is generated though, so you'd need to update the IP here: https://github.com/matrix-org/matrix-authentication-service/blob/19fab3c5fac9860c1fb41143a92ca20dee13b457/crates/config/src/sections/http.rs#L68

And then run `sh misc/update.sh` (or just `cargo run -p mas-config > ./docs/config.schema.json` if you don't have a full dev environment working. This only requires cargo/rust [to be installed](https://www.rust-lang.org/tools/install))

sandhose (approved):

PR #2945, opened 2024-07-02
Better support for secure database connections

### LINKED ISSUE #2799: Support for secure database connections using TLS certs
Hi matrix team and community!

I noticed the current docs don't mention ssl/tls connections for the database here:
https://matrix-org.github.io/matrix-authentication-service/reference/configuration.html#database

Is this supported? I recently tried to provide a connection uri with the following, but it didn't seem to work. here's the database section of my config.yaml:

```yaml
database:
  # mas-postgres-hostname is anonymized
  uri: "postgresql://mas@mas-postgres-hostname:5432/mas?sslmode=verify-full&sslkey=/etc/secrets/mas/tls.key&sslcert=/etc/secrets/mas/tls.crt&sslrootcert=/etc/secrets/ca/ca.crt"
```

 I finally also tried using the [env vars for postgresql specified in the docs](https://www.postgresql.org/docs/current/libpq-envars.html), `PGSSLKEY`,`PGSSLCERT`,`PGSSLROOTCERT`, in the env vars for the container, and then chopping off everything after the `sslmode` paramter in the uri, like below, but that didn't work either.
 
 ```yaml
database:
  # mas-postgres-hostname is anonymized
  uri: "postgresql://mas@mas-postgres-hostname:5432/mas?sslmode=verify-full"
```

This same connection uri works fine in another container, but it always says permission denied in the mas-cli docker container and returns this:

```logtalk
2024-05-25T12:51:39.125791Z INFO cli.run.init: mas_cli::commands::server: Connecting to the database
2024-05-25T12:51:39.127372Z ERROR cli.run.init:db.connect: mas_cli::util: error=could not connect to the database
Caused by:
0: error communicating with database: Permission denied (os error 13)
1: Permission denied (os error 13)
Error: could not connect to the database
Caused by:
0: error communicating with database: Permission denied (os error 13)
1: Permission denied (os error 13)
```

I have verified that the certs exist and are accessible when mounting the same certs into another container and using the same connection uri or connecting directly with psql, so I figured it's time to figure out if anyone else has made this container work with tls or if this is just my env.

This is the docker container I'm using:
https://github.com/matrix-org/matrix-authentication-service/pkgs/container/matrix-authentication-service/217725727?tag=sha-206d45b

To be clear, this works fine via database user/password credentials with ssl mode disabled, so the container itself is sound, I think.

Either way, it would be nice to support the following in the config.yaml:

```yaml
database:
  sslkey: ""
  sslrootcert: ""
  sslcert: ""
```

I don't see these options listed in here:
https://github.com/matrix-org/matrix-authentication-service/blob/206d45bb31e1afa1fea16d0194c107d302d0a593/crates/config/src/sections/database.rs#L70-L149

so perhaps it has to be added? I'm not great at rust, so if others are, this would be a huge benefit for users whose orgs require mutual TLS between all services, including databases.

sandhose:
Fixes #2799

This adds new configuration parameters to set the CA, client certificate and key, as well as the SSL mode to use

### DIFF
--- a/crates/cli/src/commands/mod.rs
+++ b/crates/cli/src/commands/mod.rs
@@ -69,16 +69,18 @@ pub struct Options {
 impl Options {
     pub async fn run(self, figment: &Figment) -> anyhow::Result<()> {
         use Subcommand as S;
+        // We Box the futures for each subcommand so that we avoid this function being
+        // big on the stack all the time
         match self.subcommand {
-            Some(S::Config(c)) => c.run(figment).await,
-            Some(S::Database(c)) => c.run(figment).await,
-            Some(S::Server(c)) => c.run(figment).await,
-            Some(S::Worker(c)) => c.run(figment).await,
-            Some(S::Manage(c)) => c.run(figment).await,
-            Some(S::Templates(c)) => c.run(figment).await,
-            Some(S::Debug(c)) => c.run(figment).await,
-            Some(S::Doctor(c)) => c.run(figment).await,
-            None => self::server::Options::default().run(figment).await,
+            Some(S::Config(c)) => Box::pin(c.run(figment)).await,
+            Some(S::Database(c)) => Box::pin(c.run(figment)).await,
+            Some(S::Server(c)) => Box::pin(c.run(figment)).await,
+            Some(S::Worker(c)) => Box::pin(c.run(figment)).await,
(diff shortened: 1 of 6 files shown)


reivilibre (approved):
I admittedly have a slight preference for just leaving this to SQLx and having users configure it through the connection URI but on the other hand this looks fine

PR #2955, opened 2024-07-05
GraphQL API to list users with a few filters

sandhose:
This can be reviewed commit by commit, it is pretty straight forward.

It adds a new `users` API which lets administrators list users with two filters:

 - `state`, to only get active or locked users
 - `canRequestAdmin`, to only get users that have that flag set (or not)

### DIFF
--- a/crates/storage-pg/src/user/mod.rs
+++ b/crates/storage-pg/src/user/mod.rs
@@ -16,15 +16,16 @@
 //! repositories
 
 use async_trait::async_trait;
-use chrono::{DateTime, Utc};
 use mas_data_model::User;
 use mas_storage::{user::UserRepository, Clock};
 use rand::RngCore;
+use sea_query::{Expr, PostgresQueryBuilder, Query};
+use sea_query_binder::SqlxBinder;
 use sqlx::PgConnection;
 use ulid::Ulid;
 use uuid::Uuid;
 
-use crate::{tracing::ExecuteExt, DatabaseError};
+use crate::{iden::Users, pagination::QueryBuilderExt, tracing::ExecuteExt, DatabaseError};
 
 mod email;
 mod password;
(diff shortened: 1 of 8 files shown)


crates/storage-pg/src/user/mod.rs
```diff
@@ -53,16 +54,29 @@ impl<'c> PgUserRepository<'c> {
...
+    // The enum_def macro generates a public enum, which we don't want, because it
+    // triggers the missing docs warning
+    #![allow(missing_docs)]
+
+    use chrono::{DateTime, Utc};
+    use sea_query::enum_def;
+    use uuid::Uuid;
+
+    #[derive(Debug, Clone, sqlx::FromRow)]
+    #[enum_def]
```
reivilibre, line 67: can you comment what this macro does? It isn't documented upstream :/: https://docs.rs/sea-query/latest/sea_query/attr.enum_def.html

It doesn't even make intuitive sense, this is a struct not an enum...?

frontend/src/gql/graphql.ts
```diff
@@ -1212,6 +1229,27 @@ export type UserAgent = {
...
+};
+
+/** An edge in a connection. */
+export type UserEdge = {
+  __typename?: 'UserEdge';
+  /** A cursor for use in pagination */
+  cursor: Scalars['String']['output'];
+  /** The item at the end of the edge */
+  node: User;
+};
```
reivilibre, line 1251: I don't understand what these are

reivilibre (approved):
seems fine, would be happy to have explanations for a couple of points though :)

crates/storage-pg/src/user/mod.rs
sandhose: This is documented here: https://docs.rs/sea-query/latest/sea_query/index.html#iden

Basically it creates an enumeration with 'names' for each field, so that you get in this case `UserLookupIden::UserId`, `UserLookupIden::Username`, etc.

This is used then in the query builder later to make sure we reference actual columns in this struct

frontend/src/gql/graphql.ts
sandhose: All the paginated lists are like this in the GraphQL API. This is following the 'Connection' convention from Relay: https://graphql.org/learn/pagination/#complete-connection-model

Basically it normalises how you do paginated lists, where each item (= edge) has a cursor, and you can paginate before and after a cursor

PR #2965, opened 2024-07-09
Backend work to support minimum password complexity

reivilibre:
Part of #172 

* Config option for minimum password complexity, default is 3 (> 10^6 guesses)
* GraphQL API exposes the configured minimum password complexity
* Minimum password complexity is enforced when setting or changing a password, but the frontend still needs work to make this interactive and pretty.

### DIFF
--- a/crates/cli/src/commands/manage.rs
+++ b/crates/cli/src/commands/manage.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::collections::BTreeMap;
+use std::{collections::BTreeMap, process::ExitCode};
 
 use anyhow::Context;
 use clap::{ArgAction, CommandFactory, Parser};
@@ -174,7 +174,7 @@ enum Subcommand {
 
 impl Options {
     #[allow(clippy::too_many_lines)]
-    pub async fn run(self, figment: &Figment) -> anyhow::Result<()> {
+    pub async fn run(self, figment: &Figment) -> anyhow::Result<ExitCode> {
         use Subcommand as SC;
         let clock = SystemClock::default();
         // XXX: we should disallow SeedableRng::from_entropy
@@ -205,7 +205,7 @@ impl Options {
 
                 if !ignore_complexity && !password_manager.is_password_complex_enough(&password)? {
                     error!("That password is too weak.");
-                    return Ok(());
+                    return Ok(ExitCode::from(1));
                 }
 
                 let password = password.into_bytes().into();
@@ -219,7 +219,7 @@ impl Options {
                 info!(%user.id, %user.username, "Password changed");
                 repo.into_inner().commit().await?;
 
-                Ok(())
+                Ok(ExitCode::SUCCESS)
             }
 
             SC::VerifyEmail { username, email } => {
@@ -256,7 +256,7 @@ impl Options {
                 repo.into_inner().commit().await?;
                 info!(?email, "Email marked as verified");
 
-                Ok(())
+                Ok(ExitCode::SUCCESS)
             }
 
             SC::IssueCompatibilityToken {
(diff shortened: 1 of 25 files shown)


crates/cli/src/commands/manage.rs
```diff
@@ -188,6 +203,11 @@ impl Options {
                     .await?
                     .context("User not found")?;
 
+                if !ignore_complexity && !password_manager.is_password_complex_enough(&password)? {
+                    error!("That password is too weak.");
+                    return Ok(ExitCode::from(1));
```
sandhose, line 208: What's the benefit compared to just throwing here? e.g. 
```suggestion
                    anyhow::bail!("That password is too weak.");
```

This would also print the error AFAIK

crates/handlers/src/views/recovery/finish.rs
```diff
@@ -226,6 +226,18 @@ pub(crate) async fn post(
             );
     }
 
+    if !password_manager.is_password_complex_enough(&form.new_password)? {
+        // TODO This error should be localised,
+        //      but actually this entire form should be pushed down into the
+        //      React frontend so user gets real-time feedback.
```
sandhose, line 232: We should indeed make a new `FieldError::Password { entropy: zxcvbn::Entropy }` enum variant? So that in the templates we can extract the entropy score, the guess times, the feedback suggestions and warnings?

This does make me think though that `is_password_complex_enough(&str) -> bool` isn't the right API, but I'm fine with leaving it like that for now

crates/config/src/sections/passwords.rs
```diff
@@ -35,6 +35,10 @@ fn default_enabled() -> bool {
     true
 }
 
+fn default_minimum_complexity() -> u8 {
+    3
```
sandhose, line 39: This might be too disruptive to existing deployments? Although it's a good default, so we might just have to make a big warning in the upgrade/release notes I guess 🤷

crates/handlers/src/test_utils.rs
```diff
@@ -178,7 +179,7 @@ impl TestState {
         let metadata_cache = MetadataCache::new();
 
         let password_manager = if site_config.password_login_enabled {
-            PasswordManager::new([(1, Hasher::argon2id(None))])?
+            PasswordManager::new(3, [(1, Hasher::argon2id(None))])?
```
sandhose, line 182: This should grab it from the site_config

sandhose (changes requested):
This is great

crates/handlers/src/views/recovery/finish.rs
reivilibre: I think the goal would be to get all that stuff client-side, because an interactive form will be much nicer than getting that from the backend. Can't see any drawback really?

crates/cli/src/commands/manage.rs
reivilibre: I kind of thought we would print the backtrace for anyhow errors. But that doesn't actually seem to be the case. So I guess we could do it that way if you prefer (but also, isn't it weird not to print the backtrace from anyhow?)

crates/config/src/sections/passwords.rs
reivilibre: It's not like it will stop anyone from logging in, so I'm not sure I share the concern to be honest. But sure, a warning in the release notes is always the right way to do it

crates/cli/src/commands/manage.rs
sandhose: It doesn't show a backtrace, it prints the Error type using `Debug`: https://doc.rust-lang.org/stable/src/std/process.rs.html#2439-2449

Which for anyhow means showing the error chain and the backtrace only if the feature is enabled (which afaik isn't the case): https://docs.rs/anyhow/latest/src/anyhow/fmt.rs.html#20-66

Anyway I'm fine with using ExitCode now

crates/handlers/src/views/recovery/finish.rs
sandhose: I guess it depends how much we want the javascript-less flows to behave? Anyway, fine for now 👍

PR #2970, opened 2024-07-12
Allow reactivating users on the homeserver

### LINKED ISSUE #2101: Admin API to unlock a user
If a user has been locked with the `lockUser()` GraphQL mutation (from https://github.com/matrix-org/matrix-authentication-service/issues/1554) then it should be possible to call `unlockUser()` to unlock them.

This call needs to not work if the user was deactivated on the HS, see #2375

### LINKED ISSUE #2375: Expose whether a user is permanently deactivated or not
Either by recording the deactivation state on the user, or by querying the Synapse deactivation status.

If we do the former, we need to import that state somehow, maybe with a CLI tool.

sandhose:
- **Add a way to reactivate users on the homeserver**
- **GraphQL API to unlock a user**
- **Show whether the user is deactivated on the homeserver in the GraphQL API**

Fixes #2101
Fixes #2375

### DIFF
--- a/crates/cli/src/commands/manage.rs
+++ b/crates/cli/src/commands/manage.rs
@@ -27,7 +27,9 @@ use mas_matrix::HomeserverConnection;
 use mas_matrix_synapse::SynapseConnection;
 use mas_storage::{
     compat::{CompatAccessTokenRepository, CompatSessionRepository},
-    job::{DeactivateUserJob, JobRepositoryExt, ProvisionUserJob, SyncDevicesJob},
+    job::{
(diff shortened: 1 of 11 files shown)


reivilibre (approved):
the only thing to comment on, is I wish we could make it clearer that unlock implies reactivation.

This gets super confusing given Synapse's own notion of user locks, user deactivations and user suspensions.
It'd be really good to straighten this out for the clarity of our code, interface and documentation.
Currently, I don't think I'd enjoy being a user of these features in combination because it doesn't make sense if you don't already know all the aspects.
However I am struggling to think of anything concrete to suggest...

sandhose:
I agree with you on this. I opened an issue to track this: https://github.com/matrix-org/matrix-authentication-service/issues/2978

PR #2971, opened 2024-07-12
Fully sync the devices with the homeserver

### LINKED ISSUE #2626: Devices can get out of sync between Synapse and MAS
The provisioning and deletion of devices is not exactly reliable (see #2625), which means sometimes the device list in Synapse gets out of sync with the one in MAS.

The long term solution may be to correctly model Matrix devices in MAS instead of relying solely on the scope for OAuth sessions and on a field in the compatibility sessions.

In the meantime, we should at least have a way to force a re-sync of the whole device list for a user, and maybe even regularly do that for all users.

sandhose:
Fixes #2626

When we end session, instead of imperatively delete devices, we now sync the device list with Synapse.
This means that if two sessions have the same device, it will not wrongly delete the device.
It also means it's less likely for the device list to get out of sync.

This also makes all the devices creation synchronous, to make sure they exist on Synapse before we start using the new sessions.

This also uses per-user PG advisory locks around device syncing, to make sure we don't have race conditions when syncing those

<details><summary>Old rant about locking kept for posterity</summary>

**One thing which I'm very not sure about** is that there is potential races between the device creations and the device sync.
If there is an ongoing sync job, the sync job might build a list of devices from the database that doesn't have a new device. That device may already have been created from somewhere else on the homeserver, but haven't been committed yet to the database.
If this is the case, the device syncing job will delete from the homeserver the newly added device.

This is very unlikely to happen though, as we would need the user to login and logout at the same time. I'm not sure what the best way would be to make sure it never happens. A few possibilities I have in mind:

 - use a PG advisory lock around any device list modification
 - make device syncing/deletion only consider devices that were recently deleted (looking at the finished_at timestamp)
 - put the devices to delete in a staged zone, so that they get deleted a bit later on (30s-1min), confirming that they are indeed not in the database anymore
 - rework the job queue, so that we can schedule jobs in the future (like in 30s) and dedupe them. This way, any modification to the device list would trigger a full sync a little later, and those sync would happen only if there was no sync for that user in the last 30s
 - revert the 'synchronous' part of device creation, so that we do them through device syncs, which, assuming jobs are run one after another, should mean we don't have devices flipping from one state to another

</details>

### DIFF
--- a/crates/cli/src/commands/manage.rs
+++ b/crates/cli/src/commands/manage.rs
@@ -27,7 +27,7 @@ use mas_matrix::HomeserverConnection;
 use mas_matrix_synapse::SynapseConnection;
 use mas_storage::{
     compat::{CompatAccessTokenRepository, CompatSessionRepository},
-    job::{DeactivateUserJob, DeleteDeviceJob, JobRepositoryExt, ProvisionUserJob},
+    job::{DeactivateUserJob, JobRepositoryExt, ProvisionUserJob, SyncDevicesJob},
     user::{UserEmailRepository, UserPasswordRepository, UserRepository},
     Clock, RepositoryAccess, SystemClock,
 };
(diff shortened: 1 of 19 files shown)


reivilibre (approved):
This all seems reasonable.

I'm not super convinced about the pagination given we have to fetch all the rows anyway, I guess I'd normally fetch the rows in a streaming style but maybe that's just hard to do here?

PR #2977, opened 2024-07-16
Mark sessions as finished in bulk

sandhose:
This is reviewable commit by commit.

On top of #2970 

- **Batch finish compatibility sessions**
- **Batch finish OAuth 2.0 sessions**
- **Batch finish browser sessions**
- **Make `mas-cli manage kill-sessions` finish sessions in bulk**
- **Kill all sessions when deactivating a user**

This makes it so that when we finish sessions (like when deactivating a user), we can do that in bulk.
This introduces new APIs on the three repositories, which are then used by the `mas-cli manage kill-sessions` command.
The reason we need this, is so that we can kill all sessions when a user is deactivated, so that they don't leave active sessions behind, and don't get magically logged in once we reactivate them.


crates/storage-pg/src/compat/session.rs
```diff
@@ -341,6 +341,64 @@ impl<'c> CompatSessionRepository for PgCompatSessionRepository<'c> {
...
+    async fn finish_bulk(
+        &mut self,
+        clock: &dyn Clock,
+        filter: CompatSessionFilter<'_>,
+    ) -> Result<usize, Self::Error> {
+        let finished_at = clock.now();
+        let (sql, arguments) = Query::update()
+            .table(CompatSessions::Table)
+            .value(CompatSessions::FinishedAt, finished_at)
+            .and_where_option(filter.user().map(|user| {
```
reivilibre, line 359: should the `CompatSessionFilter` be responsible for constructing its own `WHERE` clauses? I am a bit concerned we'd forget to update this code here if we added more to the filter

reivilibre (approved):
just one comment really (but applies to all filters), rest looks sane

crates/storage-pg/src/compat/session.rs
sandhose: I did think of this, but it was annoying in particular for the compatibility sessions, as when we query, we `JOIN` on the `compat_sso_logins` and filter on that, which we can't do for updating the `finished_at` field.

I'll have a shot though in another PR

PR #2979, opened 2024-07-16
Deduplicate the query bulding logic for all dynamic filters

sandhose:
This can be reviewd commit by commit.

- **Add a trait to help building dynamic filters**
- **Use dynamic filters on compatibility sessions**
- **Use dynamic filters on compatibility SSO logins**
- **Use dynamic filters on OAuth 2.0 sessions**
- **Use dynamic filters on upstream OAuth 2.0 links**
- **Use dynamic filters on upstream OAuth 2.0 providers**
- **Use dynamic filters on user emails**
- **Use dynamic filters on users**
- **Use dynamic filters on browser sessions**
- **Add new filters on the OAuth and compat sessions**
- **Use dynamic filters on app sessions by reusing the OAuth/compat sessions filters**

This deduplicates the filtering logic for `list()` and `count()` methods (as well as `finish_bulk()` on sessions)

The most special one is the "app sessions" repository, as it does a SELECT UNION on both the OAuth and compat sessions tables, so I did some logic to split an AppSessionFilter into a OAuthSessionFilter and a CompatSessionFilter.
I had to add a few new filters to the OAuthSessionFilter and CompatSessionFilter, but that's it.

I'm fairly confident that most of those filters have associated tests, and most of it was copy-pasting the existing where clauses.


reivilibre (approved):
Thanks for doing that, it's still verbose but at least it's all contained, it feels like a good abstraction to have!

PR #2982, opened 2024-07-18
Support using recovery tickets in the GraphQL SetPassword mutation

reivilibre:
Part of #172 (in a roundabout way) — we want to push the account recovery form into the React frontend code so it can share the code for the preview of password complexity.

### DIFF
--- a/crates/handlers/src/graphql/mutations/user.rs
+++ b/crates/handlers/src/graphql/mutations/user.rs
@@ -24,7 +24,7 @@ use zeroize::Zeroizing;
 use crate::graphql::{
     model::{NodeType, User},
     state::ContextExt,
-    UserId,
+    Requester, UserId,
 };
 
 #[derive(Default)]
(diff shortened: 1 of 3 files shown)


crates/handlers/src/graphql/mutations/user.rs
```diff
@@ -542,7 +567,10 @@ impl UserMutations {
         let user_id = NodeType::User.extract_ulid(&input.user_id)?;
```
sandhose, line 567: Isn't that fundamentally a problem with this mutation? When you have a recovery ticket, you don't know the user_id; in a sense, the recovery ticket has the user ID linked to it?

This means that either we need to make the user_id optional, or we need to make a separate mutation which only takes a recovery ticket

reivilibre: Hmm. We could include the user ID in the link that's in the e-mail? But I guess that's redundant, so making the user ID optional for the recovery ticket might make sense then. I am not sure if this will turn out better or worse than a separate mutation, I don't want to duplicate the checks at least.

Issue #2219, opened 2023-12-21
Mark inactive sessions


pmaier1:
As a user I want to know about inactive sessions (no activity since > 90 days) for that I can disconnect them and keep my account safe.

PR #2983, opened 2024-07-18
Add a filter for inactive sessions

sandhose:
This is reviewable commit by commit:

- **Add an index on the last_active_at column of the sessions tables**
- **storage: add a filter by last active time on browser sessions**
- **storage: add a filter by last active time on compatibility sessions**
- **storage: add a filter by last active time on OAuth 2.0 sessions**
- **storage: add a filter by last active time on app sessions**
- **graphql: allow filtering of sessions by last activity**
- **frontend: add filter for inactive sessions**

This adds a filter to only show sessions that have not been active for more than 90 days.

It also reworks a bit the session list to make it clearer that the browser session thing is separate.
On the frontend side, it isn't exactly what's designed, because:

 - the menu component in compound is lacking
 - we can't really merge the 'browser' sessions and the 'app' sessions

image


reivilibre:
>  On the frontend side, it isn't exactly what's designed, because:

just for ref, do you have a link to the designs / an issue, in case we need to look back on this in the future?

crates/handlers/src/graphql/mod.rs
```diff
@@ -501,3 +502,13 @@ where
         session.map(Into::into).unwrap_or_default()
     }
 }
+
+/// A filter for dates, with a lower bound and an upper bound
+#[derive(InputObject, Default, Clone, Copy)]
+pub struct DateFilter {
+    /// The lower bound of the date range
```
reivilibre, line 509: I think this is what gets generated as the GraphQL documentation right?

I think we should describe the format of the date here, since it's not obvious how a caller should use it.

I then wonder how much we care about timezones. Possibly not at all, given 90 days is far enough away that it probably doesn't matter whether it's a few hours off.

reivilibre (approved):
this seems fine to me. Glad we split up the filters like that

crates/handlers/src/graphql/mod.rs
sandhose: The scalar used is a custom one, which is serialises as a string, and has the following (autogenerated) documentation on it:

```graphql
"""
Implement the DateTime<Utc> scalar

The input/output is a string in RFC3339 format.
"""
scalar DateTime
```

so that's probably fine?

sandhose:
> just for ref, do you have a link to the designs / an issue, in case we need to look back on this in the future?

You're right, for the 'feature' that would be https://github.com/matrix-org/matrix-authentication-service/issues/2219

The designs are here: https://www.figma.com/design/HUysJAhv6cK6p1Pc81Fxaa/Matrix-Authentication-Service-(MAS)?node-id=4352-32522&t=lqvaJRtBgVVCC7Ix-4

PR #2985, opened 2024-07-19
Rework assets loading to fix splitting CSS chunks

sandhose:
So I had a re-read of the [Vite documentation](https://vitejs.dev/guide/backend-integration), and applied exactly what they suggest for including assets.

This fixes CSS code splitting, so we can now share most of the CSS between the server-side rendered templates and the SPA


frontend/index.html
```diff
@@ -26,19 +26,6 @@
...
-          if (list.matches) {
-            document.documentElement.classList.add("cpd-theme-dark");
-          } else {
-            document.documentElement.classList.remove("cpd-theme-dark");
-          }
-        }
-
-        query.addEventListener("change", handleChange);
-        handleChange(query);
-      })();
```
reivilibre, line 41: do we not support dark colour themes anymore? Or is this handled some other way?

reivilibre (approved):
admittedly I haven't probed hard to review this, but I'm happy to work on the 'if you tested it works, it works' principle for this.

Would like to understand what is changing about dark mode though

frontend/index.html
sandhose: We do, it's me who misunderstood how the design tokens worked at some point. It defaults to media queries, but you can override the theme by setting the `cpd-theme-dark` or `cpd-theme-light` classes.

So this piece of code isn't needed

PR #2986, opened 2024-07-19
graphql: Add a SetPasswordByRecovery mutation to perform account recovery

reivilibre:
Replaces #2982 and instead adds a whole new mutation for the account recovery case. I was sceptical at first but think this is actually fairly fine.


sandhose (approved):
That looks alright, thanks!

Issue #2240, opened 2024-01-06
Separate user and admin APIs


dklimpel:
If I have understood correctly, the graphql API is used by the clients as well as for the administration of MAS itself.

Currently you can't separate admin and client APIs cleanly to avoid exposing admin access to the internet, right?

Administrative access via internet should be prevented via IP filters or similar.
Probably two different graphql entry points are required for this.

sandhose:
We now have a new admin API, which is has its own resource, which can be mounted on a separate listener.

In addition, the GraphQL API is by default only available to the frontend, not for admin purposes, but can still be enabled on a separate listener with the `undocumented_oauth2_access` option set to `true` for admin purposes, for operations that the new admin API doesn't yet have.

PR #2993, opened 2024-07-24
Bootstrap a new REST-like admin API

sandhose:
I made this reviewable commit-by-commit.

Related to #2240

This adds a new admin API, served under `/api/admin/v1`.
There is an auto-generated OpenAPI spec which is available both in the docs and served under `/api/spec.json`.

I tried to make the API responses follow the JSON API spec, so that we have a consistent API design.

As a first step, I implemented a simple `GET /api/admin/v1/users` endpoint which returns a list of users, a `GET /api/admin/v1/users/{id}` endpoint which returns a single user by its ID, and a `GET /api/admin/v1/users/by-username/{username}` endpoint which returns a single user by its username.

The whole API is currently gated behind the `urn:mas:admin` scope.

Follow up things to do:

 - host/generate a documentation, either in-process or on the hosted docs site
 - document how to enable the admin API and use it
 - implement more endpoints
 - add tests
 - introduce more fine-grained scopes

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,6 +85,41 @@ dependencies = [
 ]
 
 [[package]]
+name = "aide"
+version = "0.13.4"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7b0e3b97a21e41ec5c19bfd9b4fc1f7086be104f8b988681230247ffc91cc8ed"
+dependencies = [
+ "aide-macros",
+ "axum",
+ "axum-extra",
+ "bytes",
+ "cfg-if",
+ "http",
+ "indexmap 2.2.6",
+ "schemars",
+ "serde",
+ "serde_json",
+ "serde_qs",
+ "thiserror",
+ "tower-layer",
+ "tower-service",
+ "tracing",
+]
+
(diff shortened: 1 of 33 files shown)


reivilibre (approved):
Paging through it, it seems like a lot of scaffolding here, but I also can't say that's wrong — Synapse could probably use more of it so the APIs are more foolproof.

10 feels like a small default page size IMO but whatever, as long as it's only the default.

Perhaps we should add a limit on the page size. It can be quite generous given this is an admin API but it feels like it'd be the right thing to have at least a minimal guard rail against operator error bringing the service down.
However if you're not in the mood for adding that now then I wouldn't say that blocks this.

sandhose:
> Perhaps we should add a limit on the page size. It can be quite generous given this is an admin API but it feels like it'd be the right thing to have at least a minimal guard rail against operator error bringing the service down.
However if you're not in the mood for adding that now then I wouldn't say that blocks this.

I would argue that if you're messing with the admin API, you're on your own in terms of server performance?
Also we don't enforce anything like that in the GraphQL API either, which is probably scarier 🙈

PR #2994, opened 2024-07-24
frontend: refactor password change form to extract double-input password creation inputs as new component

reivilibre:
This is sort of part of #172: by doing this refactoring, we can more easily support the same checks and metering in other forms.


frontend/src/components/PasswordCreationDoubleInput.tsx
```diff
@@ -0,0 +1,164 @@
...
+    passwordStrengthTint = ["red", "red", "orange", "lime", "green"][
+      passwordComplexity.score
+    ] as "red" | "orange" | "lime" | "green" | undefined;
+  }
+
+  return (
+    <>
+      <Form.Field name="new_password">
+        <Form.Label>
+          {t("frontend.password_change.new_password_label")}
```
sandhose, line 89: I guess keeping `password_change` as prefix for the password recovery makes sense. It won't when/if we start having this on the registration page; the label will probably have to be different

frontend/src/components/PasswordCreationDoubleInput.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  return result;
+};
+
+export default function PasswordCreationDoubleInput({
+  siteConfig,
+  forceShowNewPasswordInvalid,
+}: {
+  siteConfig: FragmentType<typeof CONFIG_FRAGMENT>;
+  forceShowNewPasswordInvalid: boolean;
```
sandhose, line 63: That's a long prop name, but I'm fine with that :)

sandhose (approved):

frontend/src/components/PasswordCreationDoubleInput.tsx
reivilibre: ah true, that's a pity, but probably not too bad

Issue #2541, opened 2024-03-22
Rate-limit password-based logins and registrations


sandhose:
Questions to answer:

 - what algorithm we should use?
 - should the rate limits be shared with all the MAS instances (e.g. Redis-based rate limits?) or can be instance-local?
 - should they be scoped to IPs, IP ranges, global?
 - what good defaults are?

H-Shay:
Rate limiting isn't my area of expertise but my first instinct is to look at what Synapse currently does and use that as a baseline:

It looks like the algorithm used is "leaky bucket as a meter" (https://github.com/element-hq/synapse/blob/d221512498f7e3267916a289dd2ef4f3e00728e8/synapse/api/ratelimiting.py#L40)
	
As for keys, synapse rate limits against several keys: ip address, account, and failed attempts. The rate limits are configurable and the defaults are:
```		
address:
    per_second: 0.15
    burst_count: 5
account:
    per_second: 0.18
    burst_count: 4
failed_attempts:
    per_second: 0.19
    burst_count: 7
```

These limits are in-process/non-persistent, and from what I can glean from the code the rate limits appear to be applied per worker, but I would just double-check in the backend room that I am not mistaken about that.  
As for config options, the ones listed above are the only ones I see related to rate limiting logins - I also don't see any options related to ignoring ip ranges for login (although there are ip block/allowlists for federation and url previews), I wonder if traditionally that has been handled by the load balancer or something similar. I'd also definitely ask in the security lobby if they have any insight on this issue - I am not sure how much the MAS usecase differs from Synapse, but I suspect they might have more insight/thoughs on this.

davidegirardi:
I'd take some inspiration from https://owasp.org/www-community/controls/Blocking_Brute_Force_Attacks and then here are some suggestions:
- When the application keeps state (number of logins attempts per IP address for example) make sure keeping that state does not cost more to you than the attacker. Otherwise it opens itself up for denial of service attacks.
- Consider implementing an incremental timeout between attempts, for example by doubling a specific value every time. After the first failed login, you have to wait 1 second, then 2, then 4, 8, 16, etc. This will be a minor inconvenience for legitimate users but will hinder automated attacks.
- The process above should not lock resources or consume synchronous CPU time.
- A CAPTCHA could also be an alternative.
- Make sure all the limiting is implemented server-side.

sandhose:
I had a first shot at it in #3013

[...]

PR #3013, opened 2024-07-25
Rate-limit password-based login attempts

sandhose:
This is a first step for #2541.

It sets up a `Limiter` utility, which will hold all the rate-limiting logic.
As a first candidate, I rate-limited password-based logins.

There are two levels of rate-limiting:

 - per-IP address, which has a low limit (3 per minute)
 - per-user, which has a higher limit (1800 per hour)

I'm hesitant about the latter, but this is to try avoid distributed password-guessing attacks.
You'd need about 600 IPs to effectively lock out a user, which should be fine?


crates/handlers/src/rate_limit.rs
```diff
@@ -0,0 +1,183 @@
...
+// limitations under the License.
+
+use std::{net::IpAddr, sync::Arc};
+
+use governor::{clock::QuantaClock, state::keyed::DashMapStateStore, Quota, RateLimiter};
+use mas_data_model::User;
+use nonzero_ext::nonzero;
+use ulid::Ulid;
+
+const PASSWORD_CHECK_FOR_REQUESTER_QUOTA: Quota = Quota::per_minute(nonzero!(3u32));
```
reivilibre, line 22: Not really sure either way but 3/min seems a bit harsh, just thinking about e.g. the company's shared office which has a shared IPv4.

crates/handlers/src/rate_limit.rs
```diff
@@ -0,0 +1,183 @@
...
+#[derive(Debug)]
+struct LimiterInner {
+    password_check_for_requester: KeyedRateLimiter<RequesterFingerprint>,
+    password_check_for_user: KeyedRateLimiter<Ulid>,
+}
+
+impl Default for LimiterInner {
+    fn default() -> Self {
+        Self {
+            password_check_for_requester: RateLimiter::keyed(PASSWORD_CHECK_FOR_REQUESTER_QUOTA),
```
reivilibre, line 79: can't remember for sure but I think you want to call `retain_recent` on these ratelimiters occasionally to perform housekeeping.

reivilibre (approved):
overall this seems right to me!

crates/handlers/src/compat/login.rs
```diff
@@ -650,6 +670,57 @@ mod tests {
         assert_eq!(body, old_body);
     }
 
+    /// Test that password logins are rate limited.
+    #[sqlx::test(migrator = "mas_storage_pg::MIGRATOR")]
```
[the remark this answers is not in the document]
sandhose: It's nice, it creates a temporary database and run the migrator on it, so each test gets an isolated database

crates/handlers/src/rate_limit.rs
sandhose: I'll do config and tweak those numbers in another PR

crates/handlers/src/rate_limit.rs
sandhose: Good call, I did that in 5b1c9b45

sandhose:
@reivilibre if you could glimpse at 5b1c9b45 and retick, that would be great

reivilibre (approved):

PR #3016, opened 2024-07-25
Account Recovery: replace the server-side rendered form with an interactive React-based form

reivilibre:
Part of #172: the interactive form has a live-feedback meter for password complexity

TODO (would like your help please):
- [ ] how do we redirect to the login form (non-React) after success? Maybe the right answer is just using vanilla JS navigation but would like your thoughts.

commit by commit :-)


frontend/src/routes/password.recovery.index.lazy.tsx
```diff
@@ -0,0 +1,152 @@
...
+      throw new Error("passwords mismatch; this should be checked by the form");
+    }
+
+    const response = await changePassword({ ticket, newPassword });
+
+    if (
+      response.data?.setPasswordByRecovery.status === SetPasswordStatus.Allowed
+    ) {
+      // TODO HOW DO WE NAVIGATE
+      router.navigate({ to: "/../login" });
```
sandhose, line 90: Honestly `window.location.href = '/login'` might be fine. Part of the annoying thing here is getting the exact app root, as MAS might be mounted on a subpath.

Maybe the best we can do is load the react app root (aka `/account`) but with a full page load, so something like

```ts
const router = useRouter(); // not in the if, because of the rules of hooks

const location = router.buildLocation({ to: "/" });
window.location.href = location.href;
```

sandhose (approved):

frontend/src/routes/password.recovery.index.lazy.tsx
reivilibre: Yup that works

frontend/locales/en.json
```diff
@@ -6,7 +6,8 @@
     "close": "Close",
     "continue": "Continue",
     "edit": "Edit",
-    "save": "Save"
+    "save": "Save",
+    "save_and_continue": "Save and continue"
```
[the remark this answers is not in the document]
t3chguy: if the text exists in Figma and seems a short reusable action such as this then I think its the right way to do it

PR #3017, opened 2024-07-26
graphql: Expose CAPTCHA config and whether password registration is enabled

reivilibre:
Part of #172 as I am working towards converting the registration form to React.

### DIFF
--- a/crates/handlers/src/graphql/model/site_config.rs
+++ b/crates/handlers/src/graphql/model/site_config.rs
@@ -14,15 +14,19 @@
 
 #![allow(clippy::str_to_string)] // ComplexObject macro uses &str.to_string()
 
-use async_graphql::{ComplexObject, SimpleObject, ID};
+use async_graphql::{ComplexObject, Enum, SimpleObject, ID};
 use url::Url;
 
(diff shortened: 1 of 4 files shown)


crates/handlers/src/graphql/model/site_config.rs
```diff
@@ -47,12 +51,33 @@ pub struct SiteConfig {
...
+    pub site_key: String,
+}
+
+/// Which Captcha service is being used
+#[derive(Enum, Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CaptchaService {
+    RecaptchaV2,
+    CloudflareTurnstile,
+    HCaptcha,
+}
```
sandhose, line 79: I wondered whether this should be an enum with the site_key in each variant, so that we can extend later with provider-specific options and in the frontend we'd have to match on the `__typename` instead

but because this will never be actual public API, that won't be hard to change

sandhose (approved):

crates/handlers/src/graphql/model/site_config.rs
reivilibre: had the same line of thought

PR #3019, opened 2024-07-26
Add admin API to create users

sandhose:


crates/handlers/src/admin/v1/users/add.rs
```diff
@@ -0,0 +1,324 @@
...
+        let status = match self {
+            Self::Internal(_) | Self::Homeserver(_) => StatusCode::INTERNAL_SERVER_ERROR,
+            Self::UsernameNotValid => StatusCode::BAD_REQUEST,
+            Self::UserAlreadyExists | Self::UsernameReserved => StatusCode::CONFLICT,
+        };
+        (status, Json(error)).into_response()
+    }
+}
+
+/// # JSON payload for the `POST /api/admin/v1/users` endpoint
```
reivilibre, line 99: can we rename payload to something else here? It's already ambiguous to start with, but this is the opposite meaning of the one in graphql right?

I'd be tempted more than anything to either go for `Input` and `Output` or `Request` and `Response`

crates/handlers/src/admin/v1/users/add.rs
```diff
@@ -0,0 +1,324 @@
...
+#[schemars(rename = "AddUserPayload")]
+pub struct Payload {
+    /// The username of the user to add.
+    username: String,
+
+    /// Skip checking with the homeserver whether the username is valid.
+    ///
+    /// Use this with caution! The main reason to use this, is when a user used
+    /// by an application service needs to exist in MAS to craft special
+    /// tokens (like with admin access) for them
```
reivilibre, line 110: I.... don't understand

reivilibre (changes requested):
nothing awful ;-)

crates/handlers/src/admin/v1/users/add.rs
sandhose: Use case for this is if you want to create a user on the MAS side for a user reserved by an application service, or a user which for some reason already exists in Synapse but not in MAS

For the application service, it's useful if you need a token for a AS user which also has Synapse admin access

crates/handlers/src/admin/v1/users/add.rs
```diff
@@ -0,0 +1,324 @@
...
+}
+
+// XXX: this should be shared with the graphql handler
+fn username_valid(username: &str) -> bool {
+    if username.is_empty() || username.len() > 255 {
+        return false;
+    }
+
+    // Should not start with an underscore
+    if username.get(0..1) == Some("_") {
```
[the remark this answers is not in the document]
sandhose: I don't know how I ended up writing this :D

reivilibre (approved):

PR #3020, opened 2024-07-29
Upgrade Rust toolchain to 1.80.0

sandhose:
This also tweaks the workspace-level lints configuration to add priority
on groups

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,8 +16,9 @@ package.repository = "https://github.com/matrix-org/matrix-authentication-servic
 unsafe_code = "forbid"
 
 [workspace.lints.clippy]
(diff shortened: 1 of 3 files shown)


Cargo.toml
```diff
@@ -16,8 +16,9 @@ package.repository = "https://github.com/matrix-org/matrix-authentication-servic
 unsafe_code = "forbid"
 
 [workspace.lints.clippy]
-all = "deny"
-pedantic = "warn"
```
sandhose, line 20: Cargo started complaining about this, because both a single lint (`str_to_string`) and group lints had the same priority

reivilibre (approved):

PR #3021, opened 2024-07-29
Set password admin API

sandhose:
This adds an API to set the password of a user

### DIFF
--- a/crates/handlers/src/admin/v1/users/set_password.rs
+++ b/crates/handlers/src/admin/v1/users/set_password.rs
@@ -1,0 +1,190 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use aide::{transform::TransformOperation, NoApi, OperationIo};
+use axum::{extract::State, response::IntoResponse, Json};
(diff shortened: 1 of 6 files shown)


crates/handlers/src/admin/v1/users/set_password.rs
```diff
@@ -0,0 +1,190 @@
...
+    State(password_manager): State<PasswordManager>,
+    id: UlidPathParam,
+    Json(params): Json<Request>,
+) -> Result<StatusCode, RouteError> {
+    let user = repo
+        .user()
+        .lookup(*id)
+        .await?
+        .ok_or(RouteError::NotFound(*id))?;
+
```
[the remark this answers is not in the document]
sandhose: I was assuming that because this is the admin API, it should just accept it? Synapse [does that](https://github.com/element-hq/synapse/blob/4580e12275793da77edca7dccdf778e53ca31f4a/synapse/rest/admin/users.py#L412) (and I don't think it checks for empty passwords either)

But I'd be happy to make it check here and have a `skip_password_check` parameter or something similar

crates/handlers/src/admin/v1/users/set_password.rs
```diff
@@ -0,0 +1,270 @@
...
+    let skip_password_check = params.skip_password_check.unwrap_or(false);
+    tracing::info!(skip_password_check, "skip_password_check");
+    if !skip_password_check
+        && !password_manager
+            .is_password_complex_enough(&params.password)
+            .unwrap_or(false)
+    {
+        return Err(RouteError::PasswordTooWeak);
```
reivilibre, line 115: I think this will give a `PasswordTooWeak` error code if the password manager is disabled, which could be a bit misleading, should we have an error code just for it?

reivilibre (approved):
other than one minor thing SGTM thanks :)

crates/handlers/src/admin/v1/users/set_password.rs
sandhose: Oh you're right! Did that in [87482b4](https://github.com/matrix-org/matrix-authentication-service/pull/3021/commits/87482b422230227dfdf30fa21fcc068578a2ad0c)

PR #3030, opened 2024-07-30
Deactivate user admin API

sandhose:
This adds an admin API to deactivate users

### DIFF
--- a/crates/handlers/src/admin/v1/users/deactivate.rs
+++ b/crates/handlers/src/admin/v1/users/deactivate.rs
@@ -1,0 +1,210 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


crates/handlers/src/admin/v1/users/deactivate.rs
```diff
@@ -0,0 +1,210 @@
...
+        let body: serde_json::Value = response.json();
+
+        // The locked_at timestamp should be the same as the current time
+        assert_eq!(
+            body["data"]["attributes"]["locked_at"],
+            serde_json::json!(state.clock.now())
+        );
+
+        // It should have scheduled a deactivation job for the user
+        // XXX: we don't have a good way to look for the deactivation job
```
reivilibre, line 143: Perhaps we should return an identifier for the job. That way, scripts that need to could wait for the job to complete.

reivilibre (approved):
SGTM

crates/handlers/src/admin/v1/users/deactivate.rs
sandhose: I would like something like that if we rewrite the job system (#2785)

PR #3031, opened 2024-07-30
OAuth 2.0 sessions list and get admin APIs

sandhose:
This can be reviewed commit by commit.

- **admin: model definition for the OAuth 2.0 sessions**
- **admin: setup base for oauth2 sessions endpoints**
- **admin: list OAuth 2.0 sessions API**
- **admin: get OAuth 2.0 session API**

### DIFF
--- a/crates/handlers/src/admin/v1/oauth2_sessions/list.rs
+++ b/crates/handlers/src/admin/v1/oauth2_sessions/list.rs
@@ -285,3 +285,58 @@ pub async fn handler(
         &base,
     )))
 }
+
+#[cfg(test)]
+mod tests {
+    use hyper::{Request, StatusCode};
+    use sqlx::PgPool;
+
+    use crate::test_utils::{setup, RequestBuilderExt, ResponseExt, TestState};
+
+    #[sqlx::test(migrator = "mas_storage_pg::MIGRATOR")]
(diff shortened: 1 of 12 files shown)


crates/handlers/src/admin/v1/oauth2_sessions/list.rs
```diff
@@ -285,3 +285,58 @@ pub async fn handler(
...
+                "last_active_at": null,
+                "last_active_ip": null
+              },
+              "links": {
+                "self": "/api/admin/v1/oauth2-sessions/01FSHN9AG0MKGTBNZ16RDR3PVY"
+              }
+            }
+          ],
+          "links": {
+            "self": "/api/admin/v1/oauth2-sessions?page[first]=10",
```
reivilibre, line 335: I guess too late to ask this now, but what's the idea behind this?

reivilibre (approved):
seems fine :)

crates/handlers/src/admin/v1/oauth2_sessions/list.rs
sandhose: the [JSON API](https://jsonapi.org) spec adds `self` links wherever it can, so that you get canonical URLs of even nested stuff

PR #3033, opened 2024-07-31
Lazy load all routes in the frontend

sandhose:
This lazy-loads all routes, to trim down the main bundle as much as possible, and optimise for the first contentful paint.
This can be reviewed commit by commit.

- **vite: enable preloading and make sure we load the right dev/prod deps**: At some point, because the asset loading was faulty, I disabled pre-loading from Vite. This enables it again, so that all the dependencies of a route get loaded in parallel
- **frontend: lazy-load the password complexity calculation logic**: This is an interesting one, it lazy-loads the part which calculates the code complexity, so that we can start rendering without loading all the dictionaries
- **frontend: lazy-load all routes**: There is no real logic change in there, just that all routes become .lazy, with the loader in a file and the component in another
- **frontend: don't wait before rendering**: Tanstack Router has this default behaviour of waiting 500ms before rendering the route, so that you have a chance to stare at the loading screen. Our loading screen doesn't seem to render, so it was just adding delay on a blank screen
- **frontend: move `getNinetyDaysAgo` to a shared utility**: Just some code deduplication

One thing I noticed is that for some reason, the loaders aren't kicking in until the page components are loaded. They should load in parallel, because this adds a bit to our loading waterfall. I asked on the Tanstack Router discord to see what they think about it.

## Non-scientific measurements

This is on the Slow 4G preset

#### Before

image

#### After

image

#### Before

image

#### After

image


frontend/src/components/PasswordCreationDoubleInput.tsx
```diff
@@ -46,9 +47,11 @@ const usePasswordComplexity = (password: string): PasswordComplexity => {
...
       });
     } else {
-      estimatePasswordComplexity(deferredPassword, t).then((response) =>
-        setResult(response),
-      );
+      loadPromise
+        .then(({ estimatePasswordComplexity }) =>
+          estimatePasswordComplexity(deferredPassword, t),
+        )
+        .then((response) => setResult(response));
```
reivilibre, line 54: will this potentially 'queue up' a bunch of `.then` tasks and then run them all at once when the import loads?

e.g. it might trigger multiple times if the user started typing a password before it finishes loading? Further, if the tasks run out of order, it might set the wrong score on the form?

reivilibre (approved):

frontend/src/components/PasswordCreationDoubleInput.tsx
sandhose: It will, but I wouldn't worry too much about it? 🤷

PR #3034, opened 2024-07-31
Stabilise and document the new account management options

### LINKED ISSUE #2884: Stabilise and document the local password experimental features
We currently have the following flags in the experimental config:

 - `password_registration_enabled`: defaults to true, controls whether self-service registration is allowed
 - `email_change_allowed`: defaults to true, controls whether users can change their own emails
 - `displayname_change_allowed`: defaults to true, controls whether users can change their own displayname
 - `password_change_allowed`: defaults to true, controls whether users can change their own password
 - `account_recovery_enabled`: defaults to false, controls whether users can recover their account (change their password) via email

For all of those we should:

 - find the right place in the configuration file
 - document them

sandhose:
Fixes #2884

This also changes one default behavior: registration is now disabled by default.

### DIFF
--- a/crates/config/src/sections/account.rs
+++ b/crates/config/src/sections/account.rs
@@ -1,0 +1,99 @@
+// Copyright 2024 The Matrix.org Foundation C.I.C.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use schemars::JsonSchema;
+use serde::{Deserialize, Serialize};
+
+use crate::ConfigurationSection;
+
+const fn default_true() -> bool {
+    true
+}
+
+#[allow(clippy::trivially_copy_pass_by_ref)]
+const fn is_default_true(value: &bool) -> bool {
+    *value == default_true()
+}
+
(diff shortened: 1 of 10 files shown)


crates/config/src/sections/account.rs
```diff
@@ -0,0 +1,99 @@
...
+    ///
+    /// This should be in sync with the policy in the homeserver configuration.
+    #[serde(default = "default_true", skip_serializing_if = "is_default_true")]
+    pub displayname_change_allowed: bool,
+
+    /// Whether to enable self-service password registration. Defaults to
+    /// `false` if password authentication is enabled.
+    ///
+    /// This has no effect if password login is disabled.
+    #[serde(default = "default_false", skip_serializing_if = "is_default_true")]
```
reivilibre, line 58:
```suggestion
    #[serde(default = "default_false", skip_serializing_if = "is_default_false")]
```

crates/config/src/sections/account.rs
```diff
@@ -0,0 +1,99 @@
...
+    /// Whether users are allowed to change their passwords. Defaults to `true`.
+    ///
+    /// This has no effect if password login is disabled.
+    #[serde(default = "default_true", skip_serializing_if = "is_default_true")]
+    pub password_change_allowed: bool,
+
+    /// Whether email-based password recovery is enabled. Defaults to `false`.
+    ///
+    /// This has no effect if password login is disabled.
+    #[serde(default = "default_false", skip_serializing_if = "is_default_false")]
```
reivilibre, line 70: hmm, why disable this by default?

reivilibre (approved):
This is giving me urges to write a proc macro to handle all the shenanigans for the defaults!
But I'll resist... for now

crates/config/src/sections/account.rs
sandhose: Because it's a way to take over accounts, so you need to trust your email sending pipeline, which makes me hesitant to enable by default?

PR #3038, opened 2024-07-31
Document the admin API

### LINKED ISSUE #2239: API documenation
I am missing some docs, which listener offer which endpoints.

- [the code](https://github.com/matrix-org/matrix-authentication-service/blob/main/crates/router/src/endpoints.rs)
- [the listener](https://matrix-org.github.io/matrix-authentication-service/usage/configuration.html#httplisteners)

### LINKED ISSUE #2604: Document how to leverage the client_credentials grant to gain admin-level API access
It is possible to use the `client_crendentials` grant to gain access to the GraphQL API with admin privileges, but it is not well documented.

sandhose:
This documents the admin API, explaining how to use it, and links to the OpenAPI spec, as well as Swagger UI.

Fixes #2604
Fixes #2239

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -1,4 +1,4 @@
-# Summary 
+# Summary
 
 # Introduction
 
@@ -21,11 +21,13 @@
 
 - [Policy engine](./topics/policy.md)
 - [Authorization and sessions](./topics/authorization.md)
+- [Use the Admin API](./topics/admin-api.md)
 
 # Reference
 
 - [Configuration file reference](./reference/configuration.md)
 - [GraphQL API](./reference/graphql.md)
+- [Admin API](./api/index.html)
 - [OAuth 2.0 scopes](./reference/scopes.md)
 - [Command line tool](./reference/cli/README.md)
     - [`config`](./reference/cli/config.md)
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -26,7 +26,6 @@
 # Reference
 
 - [Configuration file reference](./reference/configuration.md)
-- [GraphQL API](./reference/graphql.md)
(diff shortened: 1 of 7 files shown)


docs/SUMMARY.md
```diff
@@ -21,11 +21,13 @@
 
 - [Policy engine](./topics/policy.md)
 - [Authorization and sessions](./topics/authorization.md)
+- [Use the Admin API](./topics/admin-api.md)
 
 # Reference
 
 - [Configuration file reference](./reference/configuration.md)
 - [GraphQL API](./reference/graphql.md)
```
hughns, line 29: We should rename the GraphQL API to be Internal GraphQL API. Some things to do:

- Move the existing GraphQL API documentation into the Development section of the documentation hierachy.
- Change all references to "GraphQL API" in the documentation to "Internal GraphQL API".
- Add a callout section at the top of the Internal GraphQL API documentation to say:
  - This API is now for internal use only
  - External usage is now deprecated and will be blocked/removed in a future release
  - The new Admin API should be used instead

docs/reference/scopes.md
hughns, line 1: The `urn:mas:graphql:*` section should reference the Internal GraphQL API and make it clear that this is for internal (to MAS) usage only.

hughns (changes requested):
As per conversation this morning, we should clarify that the GraphQL API is now intended for internal (to MAS) usage only.

docs/SUMMARY.md
hughns: I've done this myself.

docs/reference/scopes.md
hughns: I've dene this myself.

sandhose:
I rebased it on top of main instead of the merge of all pending APIs, so that this can get merged without depending on the rest

PR #3050, opened 2024-08-01
Add GraphQL mutation to do self-service user registration

reivilibre:
Part of #172 — the registration form is being migrated to React, so the current server-side rendered registration page will be removed soon.

### DIFF
--- a/crates/handlers/src/graphql/mutations/user.rs
+++ b/crates/handlers/src/graphql/mutations/user.rs
@@ -12,19 +12,25 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use std::str::FromStr;
+
 use anyhow::Context as _;
-use async_graphql::{Context, Description, Enum, InputObject, Object, ID};
+use async_graphql::{Context, Description, Enum, InputObject, Object, SimpleObject, ID};
+use lettre::Address;
 use mas_storage::{
-    job::{DeactivateUserJob, JobRepositoryExt, ProvisionUserJob},
+    job::{DeactivateUserJob, JobRepositoryExt, ProvisionUserJob, VerifyEmailJob},
     user::UserRepository,
 };
 use tracing::{info, warn};
 use zeroize::Zeroizing;
 
-use crate::graphql::{
-    model::{NodeType, User},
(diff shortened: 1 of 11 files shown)


crates/handlers/src/graphql/mutations/user.rs
```diff
@@ -766,4 +860,154 @@ impl UserMutations {
...
+            )
+            .await
+            .is_ok();
+
+        if !passed_captcha {
+            return RegisterUserStatus::InvalidCaptcha.into();
+        }
+
+        if input.username.is_empty() {
+            // TODO are there no other grammar restrictions???
```
sandhose, line 905: hmm, there is the `username_valid` function to validate locally? but then because we call `is_localpart_available`, the HS will validate that we're giving a good username

crates/handlers/src/graphql/mutations/user.rs
```diff
@@ -766,4 +860,154 @@ impl UserMutations {
...
+            return RegisterUserStatus::InvalidPassword.into();
+        }
+
+        // If the site has terms of service, the user must accept them
+        if site_config.tos_uri.is_some() && !input.accept_terms {
+            return RegisterUserStatus::MustAcceptTerms.into();
+        }
+
+        let res = policy
+            .evaluate_register(&input.username, &input.email)
```
sandhose, line 937: Also the username is validated here

crates/handlers/src/graphql/mutations/user.rs
```diff
@@ -766,4 +860,154 @@ impl UserMutations {
...
+            .schedule_job(ProvisionUserJob::new(&user))
+            .await?;
+
+        repo.save().await?;
+
+        activity_tracker
+            .record_browser_session(&clock, &session)
+            .await;
+
+        RegisterUserStatus::Allowed.into()
```
sandhose, line 1011: Not sure if possible, but this should set the session cookie, else the user won't be authenticated after this

reivilibre: not exactly proud of it, but have added a workaround for this now

PR #3090, opened 2024-08-07
Add configuration for rate-limiting of logins, replacing hardcoded limits

reivilibre:
Follow up to #3013 to make it configurable.


crates/config/src/sections/rate_limiting.rs
```diff
@@ -0,0 +1,158 @@
...
+    /// change their own password.
+    #[serde(default = "default_login_per_account")]
+    pub per_account: RateLimiterConfiguration,
+}
+
+#[derive(Copy, Clone, Debug, Serialize, Deserialize, JsonSchema, PartialEq)]
+pub struct RateLimiterConfiguration {
+    /// A one-off burst of actions that the user can perform
+    /// in one go without waiting.
+    /// Replenishes at the rate.
```
sandhose, line 57: What does that mean?

crates/config/src/sections/rate_limiting.rs
```diff
@@ -0,0 +1,158 @@
...
+    pub burst: u32,
```
sandhose, line 58: Could we just use `NonZeroU32` here? I think the generated schema for that is fine

crates/config/src/sections/rate_limiting.rs
```diff
@@ -0,0 +1,158 @@
...
+            |limiter: &RateLimiterConfiguration| -> Option<figment::error::Error> {
+                if limiter.burst == 0 {
+                    return Some(figment::error::Error::custom("`burst` must not be zero, as this would mean the action could never be performed"));
+                }
+
+                let recip = limiter.per_second.recip();
+                // period must be at least 1 nanosecond according to the governor library
+                if recip < 1.0e-9 || !recip.is_finite() {
+                    return Some(figment::error::Error::custom(
+                        "`per_second` must be a number that is more than zero and less than 1_000_000_000 (1e9)",
```
sandhose, line 93: hmm that error message doesn't look right?

sandhose (commented):
Generally looks good, thanks!

crates/config/src/sections/rate_limiting.rs
reivilibre: how so?

crates/config/src/sections/rate_limiting.rs
reivilibre: yeah mrm it's really hard to describe this somehow. But I think I will strike off this sentence, hopefully the rest is evident enough

crates/config/src/sections/rate_limiting.rs
reivilibre: actually yeah that's a bit better, the error message is acceptable too

sandhose (approved):

crates/config/src/sections/rate_limiting.rs
reivilibre: maybe I should apply the validation on the input, I just didn't want to misthink the divisions and have it fail to create a quota with the numbers.

sandhose: actually I think what you did is fine, since it's closer to what governor does internally, I just missed the `.recip()`

PR #3093, opened 2024-08-07
Add rate-limiting for account recovery and registration

### LINKED ISSUE #2541: Rate-limit password-based logins and registrations

### LINKED ISSUE #2992: Rate-limit sending of e-mails
* 'Resend email' button on account recovery form is not ratelimited
* probably we need to apply the ratelimit overall to any e-mail initiated by an anonymous user, at the very least
	* registration
	* recovery
	* resends of the above
* Probably ditto for logged in users that are trying to add a new address to their account

Users spamming themselves is still undesirable, but less awful. But worth noting this probably still impacts the sender reputation so we should control that too.

For updates like 'your password has been changed' etc, I think the best way to do that might be to roll-up multiple updates into one e-mail, if we are past the limit. This is probably something that would need to shift to a different issue.

For my own IdP project I have used the  `governor` crate and that seemed to do the business.

This kind of thing could also be done in the reverse proxy (e.g. nginx can do it) but it's probably best to have it work out of the box, given we don't want to make it hard for operators to set up.

reivilibre:
Follows #3090 

Fixes #2992 
Fixes #2541

### DIFF
--- a/crates/config/src/sections/rate_limiting.rs
+++ b/crates/config/src/sections/rate_limiting.rs
@@ -43,8 +43,8 @@ pub struct LoginRateLimitingConfig {
     ///
     /// Note: this limit also applies to password checks when a user attempts to
     /// change their own password.
-    #[serde(default = "default_login_per_address")]
-    pub per_address: RateLimiterConfiguration,
+    #[serde(default = "default_login_per_ip")]
+    pub per_ip: RateLimiterConfiguration,
     /// Controls how many login attempts are permitted
     /// based on the account that is being attempted to be logged into.
     /// This can protect against a distributed brute force attack
@@ -139,7 +139,7 @@ impl ConfigurationSection for RateLimitingConfig {
             return Err(error_on_field(error, "registration"));
         }
 
-        if let Some(error) = error_on_limiter(&self.login.per_address) {
+        if let Some(error) = error_on_limiter(&self.login.per_ip) {
             return Err(error_on_nested_field(error, "login", "per_address"));
         }
         if let Some(error) = error_on_limiter(&self.login.per_account) {
@@ -166,7 +166,7 @@ impl RateLimiterConfiguration {
     }
 }
 
-fn default_login_per_address() -> RateLimiterConfiguration {
+fn default_login_per_ip() -> RateLimiterConfiguration {
     RateLimiterConfiguration {
         burst: NonZeroU32::new(3).unwrap(),
         per_second: 3.0 / 60.0,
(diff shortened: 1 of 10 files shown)


crates/config/src/sections/rate_limiting.rs
```diff
@@ -92,7 +124,22 @@ impl ConfigurationSection for RateLimitingConfig {
...
+                "per_address",
+            ));
+        }
+
+        if let Some(error) = error_on_limiter(&self.registration) {
+            return Err(error_on_field(error, "registration"));
+        }
+
+        if let Some(error) = error_on_limiter(&self.login.per_ip) {
             return Err(error_on_nested_field(error, "login", "per_address"));
```
sandhose, line 143:
```suggestion
            return Err(error_on_nested_field(error, "login", "per_ip"));
```

crates/handlers/src/rate_limit.rs
```diff
@@ -66,15 +81,25 @@ type KeyedRateLimiter<K> = RateLimiter<K, DashMapStateStore<K>, QuantaClock>;
 
 #[derive(Debug)]
 struct LimiterInner {
+    account_recovery_per_requester: KeyedRateLimiter<RequesterFingerprint>,
+    account_recovery_per_email: KeyedRateLimiter<String>, // TODO lowercased?
```
sandhose, line 85: It could indeed have some kind of normalisation going on, but fine for now

crates/handlers/src/views/recovery/progress.rs
```diff
@@ -130,14 +132,25 @@ pub(crate) async fn post(
     // Verify the CSRF token
     let () = cookie_jar.verify_form(&clock, form)?;
 
+    // Check the rate limit if we are about to process the form
+    if let Err(e) = limiter.check_account_recovery(requester, &recovery_session.email) {
+        tracing::warn!(error = &e as &dyn std::error::Error);
```
sandhose, line 137: I really wish tracing had a sigil (like `?` or `%`) to report as `&dyn Error` :(

crates/handlers/src/rate_limit.rs
reivilibre: heh actually I already did this

sandhose (approved):

crates/handlers/src/views/recovery/progress.rs
reivilibre: https://github.com/tokio-rs/tracing/issues/1308

